PR #80, opened 2015-04-06
Projects API: Change urlencode to rawurlencode

paolomainardi:
Trying to search a project like this: "Test project" it gets converted to "Test+Project", instead RedmineAPI expect the space encoded as "%20", which is correctly handled by rawurlencode function.


m4tthumphrey:
Thanks for this @paolomainardi. I will not be merging this as you've only updated the Project endpoints but I shall make sure the library is updated to use `rawurlencode()` in the [next release](https://github.com/m4tthumphrey/php-gitlab-api/issues/81).

PR #88, opened 2015-04-29
Encode dots in project name

LukasVice:
If there are dots in the project name, the GitLab API does not work properly with urlencode only. The dots have to be replaced with %2E. See https://github.com/gitlabhq/gitlabhq/issues/8553


m4tthumphrey:
Fixed in 2b320f7

PR #110, opened 2016-01-13
Allow to access `Note` id through `__get`

shulard:
In [gushphp/gush](https://github.com/gushphp/gush), we need to access the comment `id`. The simplest way is to use the `Model` definition to map loaded data.

It's a really simple PR but it'll allow to simplify our code :smile:.


m4tthumphrey:
:+1:

shulard:
:smile: thanks for your quick approval !

PR #120, opened 2016-05-25
Add tests for php7

garak:
I just tried on my local machine and tests with php7 are running fine


m4tthumphrey:
Thanks! Ignore the 5.3.3 issue, think Travis has an issue...

PR #126, opened 2016-06-20
Added support for milestone filter

beeglebug:
As per the latest docs, a milestone title can be passed as an argument to filter the issues by that milestone (http://docs.gitlab.com/ee/api/issues.html#list-project-issues)

This adds a milestone key to the array of allowed params.


beeglebug:
Thanks. Are you planning on a new release any time soon? I am using a downstream package (http://github.com/vinkla/gitlab) which is pulling in `^7.13`, so even an incremental release would let me pull it into production.

m4tthumphrey:
Yep, just pushed 7.13.1. Didn't realise it had been nearly a year since the last release 😮

Issue #54, opened 2014-10-13
PHP Version support


ls12styler:
What should be the minimum supported version of PHP? Considering 5.3 was released over 5 years ago and that 5.6 was recently released, I would suggest the library targets no less that 5.5.

m4tthumphrey:
I am thinking of going ahead with making the library PHP 5.4+ in the next major update

- Guzzle4 requires 5.4
- It's 2015
Thoughts? @ls12styler @hollodotme @jubianchi

ls12styler:
Anything is better than 5.3 :) so thumbs up from me.

m4tthumphrey:
Wow, glad we didn't switch to Guzzle4 as Guzzle is now on 6.2!

fbourigault:
I'm closing this issue as 9.0 will ne PHP 5.6+. See #195.

PR #134, opened 2016-07-28
Remove PHP 5.3 versions from the travis build file

EdwinHoksberg:
This makes sure the builds can successfully run again.

As discussed in #54


m4tthumphrey:
5.3 was actually ok, was just 5.3.3...

EdwinHoksberg:
Yes I know, but I thought you wanted to drop 5.3 entirely? If not i'll add it back :)

m4tthumphrey:
5.3 support has come up before and whilst I agree that everyone should be on _at least_ 5.4, arguably 5.6, I don't feel it's necessary to force >= 5.4 for the sake of it. There's nothing > 5.4 that we need in this lib yet so we may as well support 5.3 for now.

Yeh if you could add 5.3 back that'd be great, can finally get rid of the `build:error` button!

EdwinHoksberg:
Yeah I agree with you, I've added it back :+1:

m4tthumphrey:
👍

PR #149, opened 2016-10-03
added remove method for Issues

brammeleman:


lib/Gitlab/Api/Issues.php
```diff
@@ -58,6 +58,16 @@ public function update($project_id, $issue_id, array $params)
      * @param int $issue_id
      * @return mixed
      */
+    public function remove($project_id, $issue_id)
```
radutopala: @brammeleman let's pls use `delete` on the method name, to keep the API's naming: https://docs.gitlab.com/ce/api/issues.html#delete-an-issue.

brammeleman: Mmmh... I'm confused: the gitlab API just tells to do a http delete on an issue id, which is like any other  object in gitlab.

brammeleman: All other objects use remove:
```sh
~/src/php-gitlab-api/lib/Gitlab/Api $ grep function * | egrep 'remove|delete'
AbstractApi.php:    protected function delete($path, array $parameters = array(), $requestHeaders = array())
Groups.php:    public function remove($group_id)
Groups.php:    public function removeMember($group_id, $user_id)
Issues.php:    public function remove($project_id, $issue_id)
Projects.php:    public function remove($project_id)
Projects.php:    public function removeMember($project_id, $user_id)
Projects.php:    public function removeHook($project_id, $hook_id)
Projects.php:    public function removeKey($project_id, $key_id)
Projects.php:    public function removeLabel($project_id, $name)
Projects.php:    public function removeForkRelation($project_id)
Projects.php:    public function removeService($project_id, $service_name)
Projects.php:    public function removeVariable($project_id, $key)
Repositories.php:    public function deleteBranch($project_id, $branch_name)
Repositories.php:    public function deleteFile($project_id, $file_path, $branch_name, $commit_message)
Snippets.php:    public function remove($project_id, $snippet_id)
SystemHooks.php:    public function remove($id)
Users.php:    public function remove($id)
Users.php:    public function removeKey($id)
Users.php:    public function removeUserKey($user_id, $key_id)
```

radutopala: Well.. there are 2 there :) .. we should rename the Repositories ones..

PR #153, opened 2016-10-21
removing bogus $params from Issues::remove. fixes #152

### LINKED ISSUE #152: Issues->remove() fails with Undefined variable: params
Issues->remove() does not accept $params from caller but passes variable to AbstractApi->delete() method.
This results in:
> PHP Catchable fatal error:  Argument 2 passed to Gitlab\Api\AbstractApi::delete() must be of the type array, null given, ...

dariuskt:


m4tthumphrey:
Well spotted!

PR #157, opened 2016-11-15
Add deploy keys to projects api

jk:
I needed the deploy keys for a given project. So I implemented it according to the [documentation](https://docs.gitlab.com/ce/api/deploy_keys.html#list-project-deploy-keys).

### DIFF
--- a/lib/Gitlab/Api/Projects.php
+++ b/lib/Gitlab/Api/Projects.php
@@ -236,6 +236,14 @@ class Projects extends AbstractApi
 
     /**
      * @param int $project_id
+     * @return mixed
+     */
+    public function deployKeys($project_id) {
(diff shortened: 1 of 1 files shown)


lib/Gitlab/Api/Projects.php
```diff
@@ -236,6 +236,14 @@ public function removeMember($project_id, $user_id)
 
     /**
      * @param int $project_id
+     * @return mixed
+     */
+    public function deployKeys($project_id) {
```
radutopala: @jk please move `{` to the next line

jk:
@radutopala done. But there are a couple more of these coding style violations in that and other files. But this is out of the scope of this PR.

PR #158, opened 2016-11-29
Implement multi-file commits

SautdeChat:
First, thanks for the API, it has been super helpful to my current project.

< backstory>
I have recently updated my GitLab and noticed an super helpful feature: Committing multiple files at once!
Until now I have been running a loop to add files to a repository. But this will create problems for runners (queuing up multiple for each single-file commit) which means in addition, I would have to remove and re-add the ci.yml.
<\backstory>

Anyway, they have added a new "[Create a commit with multiple files and actions](https://docs.gitlab.com/ce/api/commits.html#create-a-commit-with-multiple-files-and-actions)" option, and I have added a new function to Repositories.php to utilise it.
This new api call, however, requires JSON be sent, so I had to remove Your array typecasting in the httpclient. Not sure if this will upset something I am not currently using.

Here is how I am calling the new function and creating the data contents:
(Not sure if You would want to include an Action model somewhere?)

```
$files = ['install.yml', 'setup.yml'];

try {
    // Create the $actions array
    $actions = [];
    foreach ($files as $file) {
        // Make a new action class
        $action = new \stdClass();
        // Fill required properties
        $action->action = "update";
        $action->file_path = $file;
        $action->content = "Test Content";
        // Add to actions array
        array_push($actions, $action);
    }
    // Call actions function with required parameters
    // Is converted to JSON in actions()
    return GitLab::api('repo')->actions($id, $actions, 'develop', 'multipletest');
} catch (Exception $e) {
    $this->streamUpdate(0, $e->getMessage());
}
```

Let me know if I have to change anything. Or if it can be improved. =)

### DIFF
--- a/lib/Gitlab/Api/AbstractApi.php
+++ b/lib/Gitlab/Api/AbstractApi.php
@@ -60,7 +60,7 @@ abstract class AbstractApi implements ApiInterface
      * @param array $files
      * @return mixed
      */
-    protected function post($path, array $parameters = array(), $requestHeaders = array(), array $files = array())
+    protected function post($path, $parameters = array(), $requestHeaders = array(), array $files = array())
     {
         $response = $this->client->getHttpClient()->post($path, $parameters, $requestHeaders, $files);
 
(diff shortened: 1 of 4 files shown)


fbourigault:
Hello, thanks for your contribution! However, the master has evolved to use [HTTPlug](http://httplug.io/). Could you rework your PR to make it compatible?

PR #169, opened 2017-02-06
added 'addNote' method to MergeRequests API

alfredbez:
I've also reformatted the file, because there were ugly [^M](http://stackoverflow.com/questions/1822849/what-are-these-ms-that-keep-showing-up-in-my-files-in-emacs) at the end of every line.

[Clean diff](https://github.com/m4tthumphrey/php-gitlab-api/pull/169/files?w=1) (I added `?w=1` to the url)

PR #170, opened 2017-02-10
fixed wrong path in addNote method

alfredbez:
merge_request -> merge_requests

introduced in #169


alfredbez:
see also https://docs.gitlab.com/ce/api/notes.html#create-new-merge-request-note

PR #174, opened 2017-03-22
allow autor for create/update/delete

glaubinix:
File api calls allow 2 additional optional parameters: `author_email` and `author_name`
GitLab documentation for the calls: https://docs.gitlab.com/ee/api/repository_files.html


m4tthumphrey:
Thanks!

PR #175, opened 2017-03-25
Update README.md

Theaxiom:
Updating command to install package via composer. Do not edit `composer.json` or `composer.lock` directly and don't run `composer update`


m4tthumphrey:
👍

PR #187, opened 2017-05-05
Allowing different archive format types

violuke:
Allowing different archive format types. Backwards compatible as default is `tar.gz` which is what was the default before.
Correcting incorrect comment about "content" key in array too.

### DIFF
--- a/lib/Gitlab/Api/Repositories.php
+++ b/lib/Gitlab/Api/Repositories.php
@@ -337,16 +337,14 @@ class Repositories extends AbstractApi
     }
 
     /**
-     * File content is base64 encoded and placed in the "content" index of the returning array.
-     * You can then save the content with the tar.gz extension
-     *
      * @param int $project_id
      * @param array $params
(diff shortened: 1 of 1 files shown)


violuke:
Note, I don't believe that the failed test here is related to my change:
```
There was 1 failure:

1) Gitlab\Tests\Api\ProjectsTest::shouldGetOwnedProjects
Expectation failed for method name is equal to <string:get> when invoked 1 time(s)
Parameter 0 for invocation Gitlab\Api\AbstractApi::get('projects?owned=1', Array (...), Array ()) does not match expected value.
Failed asserting that two strings are equal.

--- Expected
+++ Actual
@@ @@
-'projects/owned'
+'projects?owned=1'

/home/travis/build/m4tthumphrey/php-gitlab-api/lib/Gitlab/Api/Projects.php:56
/home/travis/build/m4tthumphrey/php-gitlab-api/test/Gitlab/Tests/Api/ProjectsTest.php:70
```

PR #198, opened 2017-06-23
Update build variables api to understand the new protection function

EdwinHoksberg:
This is a new feature, introduced in Gitlab 9.3: https://about.gitlab.com/2017/06/22/gitlab-9-3-released/#protected-variables-for-enhanced-pipelines-security

One thing I'm unsure about; how is your stance on backwards compatibility? Because this change will probably break older Gitlab versions.


m1guelpf:
ping @m4tthumphrey

lib/Gitlab/Api/Projects.php
```diff
@@ -736,6 +736,19 @@ public function removeVariable($project_id, $key)
         return $this->delete($this->getProjectPath($project_id, 'variables/'.$this->encodePath($key)));
     }
 
+    /**
+     * @param int $project_id
```
glensc, line 740: why is the $project_id always typed as integer even it's allowed to use project_path as well?

EdwinHoksberg: This PR is almost three years old, I have no idea!

PR #207, opened 2017-07-18
Compatibility with v4 api for Users::block and Users::unblock

fbourigault:
Changes from https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9371.


fbourigault:
Thank you for review!

PR #211, opened 2017-07-19
remove expires_at snippet property

fbourigault:
Changes from [!8723](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8723).


fbourigault:
Thank you for the review!

PR #213, opened 2017-07-20
Projects collection consolidation

fbourigault:
This contains changes for [!8962](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8962), [!8877](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8877) and [!9674](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9674).

I introduced the `OptionsResolver` to validate and sanitize parameters. If we agree with this, the same pattern could be applied everywhere.

### DIFF
--- a/lib/Gitlab/Api/AbstractApi.php
+++ b/lib/Gitlab/Api/AbstractApi.php
@@ -6,6 +6,7 @@ use Http\Discovery\StreamFactoryDiscovery;
 use Http\Message\MultipartStream\MultipartStreamBuilder;
 use Http\Message\StreamFactory;
 use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Abstract class for Api classes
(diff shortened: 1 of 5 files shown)


lib/Gitlab/Api/Projects.php
```diff
@@ -2,77 +2,57 @@
...
-            'order_by' => $order_by,
-            'sort' => $sort
-        ));
+     * @param array $parameters
+     * @return mixed
+     */
+    public function all(array $parameters = [])
+    {
+        $resolver = $this->createOptionsResolver();
+        $normalizer = function ($value) {
```
dbu: i would call this more specific, maybe $booleanNormalizer

lib/Gitlab/Api/Projects.php
```diff
@@ -2,77 +2,57 @@
...
-     */
-    public function search($query, $page = 1, $per_page = self::PER_PAGE, $order_by = self::ORDER_BY, $sort = self::SORT)
-    {
-        return $this->get('projects/search/'.$this->encodePath($query), array(
-            'page' => $page,
-            'per_page' => $per_page,
-            'order_by' => $order_by,
-            'sort' => $sort
-        ));
+     * @param array $parameters
```
dbu: imho this should explain what parameters are valid and what they mean. something like
```
@param array $parameters
    @var bool  archived Whether to include archived projects in the result (if thats what it means)
    @var string visibility  One of public|internal|private
    ...
    (include or reference the options from the base class too)
```

the old methods had a parameter list so you had an idea of what parameters you have, but still can leave you guessing at the meaning.

lib/Gitlab/Api/AbstractApi.php
```diff
@@ -190,6 +191,29 @@ protected function encodePath($path)
         return str_replace('.', '%2E', $path);
     }
 
+    /**
+     * Create a new OptionsResolver with page, per_page and sort options.
+     *
+     * @return OptionsResolver
+     */
+    protected function createOptionsResolver()
```
dbu: why is this its own method? its only used once... or is this a first step and more methods will be refactored to this?

fbourigault: This will be used everywhere once we agreed on the current implementation.

fbourigault:
Thanks for review!

PR #218, opened 2017-07-25
replace visibility_level by visibility

fbourigault:
Changes for [!9337](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9337)


fbourigault:
Thank you for review!

PR #221, opened 2017-07-31
Update MergeRequestsApi for V4 compatibility

fbourigault:
This carry changes for [!8793](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8793).

I introduced a QueryStringBuilder class to replace `http_build_query`. The only difference is that indexed arrays are encoded using `[]` instead of `[0]`.


fbourigault:
Thank you for review!

PR #224, opened 2017-08-01
Use options resolver almost everywhere

### LINKED ISSUE #186: No option to search on a date
I noticed the Gitlab API supports a since and until date to get commits. https://docs.gitlab.com/ce/api/commits.html

The function commits in Repositories.php does not support this.

fbourigault:
This force usage of `OptionsResolver` only where it would break BC. It may be used later in other methods without breaking.

This improve a lot the API quality by avoiding endless method signatures. All the updated methods have their arguments in sync with current v4 API.

This also fix #186.

This is probably better reviewed commit per commit instead of a whole.

## To Do
* [x] Fix the models.


fbourigault:
@m1guelpf could you review this one? This is the only missing stuff before I release a beta!

fbourigault:
Thank you for the review!

PR #225, opened 2017-08-02
Static analysis

fbourigault:
This fix level 4 errors reported by https://github.com/phpstan/phpstan.

This improve code safety.

I introduced explicit methods to get API endpoints in the Client. In near future, we could deprecate the `api` and `__get` magic methods from `Gitlab\Client` which does not add any value.

phpstan is now run on travis-ci.


fbourigault:
Thank you for review!

PR #230, opened 2017-08-10
Add .gitattributes

vinkla:
[Exclude non-essential files in .gitattributes.](https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+
+* text=auto
+
+/doc export-ignore
+/tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.styleci.yml export-ignore
+/.travis.yml export-ignore
+/phpunit.xml.dist export-ignore
+/README.md export-ignore
+/UPGRADE.md export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,11 @@
+* text=auto
+
+/doc export-ignore
+/tests export-ignore
```
m1guelpf: Why are we ignoring the tests?

.gitattributes
```diff
@@ -0,0 +1,11 @@
...
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.styleci.yml export-ignore
+/.travis.yml export-ignore
+/phpunit.xml.dist export-ignore
+/README.md export-ignore
```
m1guelpf: And why are we ignoring the README?

.gitattributes
vinkla: We don't need the tests directory in production.

.gitattributes
vinkla: We don't need the README.md file in production.

PR #232, opened 2017-08-10
Add pagination to load branches

kuzaxak:
* Add pagination to loading of list branches

This patch is compatible with gitlab api v3

### DIFF
--- a/lib/Gitlab/Api/Repositories.php
+++ b/lib/Gitlab/Api/Repositories.php
@@ -4,11 +4,15 @@ class Repositories extends AbstractApi
 {
     /**
      * @param int $project_id
+     * @param array $parameters
      * @return mixed
      */
-    public function branches($project_id)
+    public function branches($project_id, array $parameters = [])
     {
-        return $this->get($this->getProjectPath($project_id, 'repository/branches'));
+        $resolver = $this->createOptionsResolver();
+        $resolver->setDefined('query');
+
+        return $this->get($this->getProjectPath($project_id, 'repository/branches'), $resolver->resolve($parameters));
     }
 
     /**


lib/Gitlab/Api/Repositories.php
```diff
@@ -4,11 +4,15 @@ class Repositories extends AbstractApi
...
      * @param int $project_id
+     * @param array $parameters
      * @return mixed
      */
-    public function branches($project_id)
+    public function branches($project_id, array $parameters = [])
     {
-        return $this->get($this->getProjectPath($project_id, 'repository/branches'));
+        $resolver = $this->createOptionsResolver();
+        $resolver->setDefined('query');
```
fbourigault: According to https://docs.gitlab.com/ee/api/branches.html#list-repository-branches, there is no `query` filter available for this branch. Could you remove it?

kuzaxak: In api v4 when you try load list of branches you receive Header with pagination. Also pagination query works. 

Note: you can compare queries.
 
V3: 
http://take.ms/e4hJW

V4:
http://take.ms/NNhQK
http://take.ms/8AD67

kuzaxak: I think it is issue related with API documentation.

blankse:
@kuzaxak Can you also add pagination for tags?

fbourigault:
Please, open an other PR for tags!

lib/Gitlab/Api/Repositories.php
fbourigault: As the filter is undocumented, I would not add it for now and try to learn more about why it's not documented.

kuzaxak: Without this fix you can't load all branches.

fbourigault: You can using the `Gitlab\ResultPager::fetchAll()`!

However, I'm not willing to add undocumented methods/parameters to this library. There is probably a reason about it not being documented. We can move forward with this PR by removing the `query` option, and add-it later if it was a Gitlab API documentation issue.

tobiasetter:
It is already possible to use pagination for branches so I think we don't need a merge here.

The only thing we could add is a test for the pagination as I added for the project members.

GrahamCampbell:
It looks like this has already been implemented.

PR #234, opened 2017-08-16
Multi files commit

### LINKED ISSUE #209: How to post more files?
I use the following method for posting more files to Gitlab: https://github.com/AntoninSlejska/php-gitlab-api/blob/8.0.1/lib/Gitlab/Api/Repositories.php#L369
The array $actions contains one or more objects: https://github.com/AntoninSlejska/php-gitlab-api/blob/8.0/docs/actions-array.txt

Is it possible to use for posting of more files some of the existing methods?

fbourigault:
This add [multi file commit](https://docs.gitlab.com/ee/api/commits.html#create-a-commit-with-multiple-files-and-actions) method.

Fix #209 and close #158.

## To do
- [x] Tests.

### DIFF
--- a/lib/Gitlab/Api/AbstractApi.php
+++ b/lib/Gitlab/Api/AbstractApi.php
@@ -89,7 +89,7 @@ abstract class AbstractApi implements ApiInterface
 
         $body = null;
         if (empty($files) && !empty($parameters)) {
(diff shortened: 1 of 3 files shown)


AntoninSlejska:
Thanks! It looks great.

fbourigault:
Thank you for the review!

PR #236, opened 2017-08-22
Update ProjectHook.php

luceos:
adds pipeline events, see https://docs.gitlab.com/ce/api/projects.html#edit-project-hook


fbourigault:
Thank you!

PR #237, opened 2017-08-25
Fix false boolean query string encoding

fbourigault:
The `QueryStringBuilder` has a bug with encoding the `false` value.

See https://3v4l.org/r4Sqt for the `http_build_query` behavior.


fbourigault:
Thank you for the review!

PR #238, opened 2017-08-25
Add pagination to load tags

blankse:
Similar to https://github.com/m4tthumphrey/php-gitlab-api/pull/232


fbourigault:
Thank you!

PR #240, opened 2017-08-28
Deprecate merge request comment related methods

fbourigault:
Comments have been removed in V4. We should use notes api instead.


fbourigault:
Thank you for the review!

Issue #242, opened 2017-09-07
Gitlab/Model/MergeRequest methods fails on incorrect id usage


KyleWiering:
Attempting to use the methods within the model fail as the model is using the mergeRequest's id, not the iid.

KyleWiering:
Re-opened to address one instance I missed on the previous pull request.

PR #243, opened 2017-09-07
#242 Gitlab/Model/MergeRequest methods fails on incorrect id usage

KyleWiering:
* Adjusted the model to use the iid instead of id.


fbourigault:
Thank you!

PR #248, opened 2017-09-18
Added missing property 'start_date' to Milestone model

happyproff:
`MIlestone` property `start_date` was missed. Documentation: https://docs.gitlab.com/ce/api/milestones.html

This property exists in both v3 and v4 APIs.


m4tthumphrey:
Thanks!

PR #250, opened 2017-09-19
Add commits

mykiwi:
https://docs.gitlab.com/ce/api/commits.html

### DIFF
--- a/lib/Gitlab/Api/Commits.php
+++ b/lib/Gitlab/Api/Commits.php
@@ -1,0 +1,43 @@
+<?php namespace Gitlab\Api;
+
+class Commits extends AbstractApi
+{
+    /**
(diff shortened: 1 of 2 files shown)


fbourigault:
Commits API is partially present in [`Gitlab\Api\Repositories`](https://github.com/m4tthumphrey/php-gitlab-api/blob/65ffd89fad93a5cc66f1c2283c35cdf631be5995/lib/Gitlab/Api/Repositories.php#L135-L268)!

However, I like the idea to have those in a separate class. Could you deprecate stuff in `Gitlab\Api\Repositories` to avoid duplicate things?

mykiwi:
Ok thanks.

Sorry but I will not have the time to do it.

PR #251, opened 2017-09-19
Updated addVariable() and updateVariable()

c33s:
updated `addVariable()` and `updateVariable()` to match current gitlab api (added `$protected `$environment_scope`)
 
- https://docs.gitlab.com/ee/api/project_level_variables.html#create-variable
- https://docs.gitlab.com/ee/api/project_level_variables.html#update-variable

### DIFF
--- a/lib/Gitlab/Api/Projects.php
+++ b/lib/Gitlab/Api/Projects.php
@@ -568,14 +568,22 @@ class Projects extends AbstractApi
      * @param string $environment_scope
      * @return mixed
      */
-    public function addVariable($project_id, $key, $value, $protected = false, $environment_scope ="*")
+    public function addVariable($project_id, $key, $value, $protected = null, $environment_scope = null)
     {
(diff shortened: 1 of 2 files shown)


c33s:
an idea also would be to do some input validation with `webmozart/assert`
```
Assert::nullOrBoolean($protected)
Assert::nullOrString($environment_scope)
```

c33s:
@m1guelpf what about the suggestion to add assert? this would add a new dependency to the code but will improve the error detection.  but it is a very stable dependency (26M downloads ). it is from @webmozart inpired by  beberlei/assert.

m1guelpf:
@c33s I feel like if we wanted to do something like assert does, we would add typehints. Also, I don't see the value of validating input in an API client, if something is wrong, the GitLab server will throw an error.

fbourigault:
We started using `OptionsResolver` to validate filters/payloads.

c33s:
anything missing from my side that you can merge this PR?

fbourigault:
Thank you!

PR #253, opened 2017-09-20
Allow use of search parameter for merge requests

velichkin:
It's possible to use search parameter for merge-requests. The only change is to add to $resolver.

### DIFF
--- a/lib/Gitlab/Api/MergeRequests.php
+++ b/lib/Gitlab/Api/MergeRequests.php
@@ -67,7 +67,8 @@ class MergeRequests extends AbstractApi
             ->setAllowedTypes('created_before', \DateTimeInterface::class)
             ->setNormalizer('created_before', $datetimeNormalizer)
         ;
-
+        $resolver->setDefined('search');
+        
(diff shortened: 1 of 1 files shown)


velichkin:
ping @m4tthumphrey

fbourigault:
`search` seems to be undocumented for merge requests. I don't like so much adding undocumented Gitlab API stuff as it may be considered as internal.

Do you have some references about the existence of this `search` parameter?

velichkin:
I've checked it and it works. But, unfortunately, I haven't found it in official API.

fbourigault:
@m4tthumphrey @m1guelpf what is your point of view about undocumented Gitlab API features support in this library?

velichkin:
Here is the discussion about this parameter - https://gitlab.com/gitlab-org/gitlab-ce/issues/38318

velichkin:
https://gitlab.com/gitlab-org/gitlab-ce/issues/38318

fbourigault:
Thank you!

Issue #254, opened 2017-09-21
Problem with updating an issue


phsteffen:
Hi Folks,

i found a little Bug in the issue-upate-function:

https://github.com/m4tthumphrey/php-gitlab-api/blob/70aeb9f412f8dc184dbd91afd75058ab8702746e/lib/Gitlab/Model/Issue.php#L94

To Update the issue the gitlab-api needs the `iid` and not the `id` of the issue.  
[Link to the api-doc](https://docs.gitlab.com/ee/api/issues.html#edit-issue)

I think it should be:  
`$data = $this->client->issues()->update($this->project->id, $this->iid, $params);`

phsteffen:
Same thing here:

https://github.com/m4tthumphrey/php-gitlab-api/blob/70aeb9f412f8dc184dbd91afd75058ab8702746e/lib/Gitlab/Model/Issue.php#L83

I think this should be work:

`$data = $this->client->issues()->show($this->project->id, $this->iid);`

m1guelpf:
@phsteffen Can you open a PR?

phsteffen:
@m1guelpf PR is open: https://github.com/m4tthumphrey/php-gitlab-api/pull/255
Thx.

PR #255, opened 2017-09-21
Issues API: Bugfixing in update and show an issue (Issue: #254)

phsteffen:
need the `iid` (not the `id`) for `show()` and `update()`.
See also:
- for show: https://docs.gitlab.com/ee/api/issues.html#single-issue
- for update: https://docs.gitlab.com/ee/api/issues.html#edit-issue


fbourigault:
Thank you!

PR #261, opened 2017-10-03
Add ext-xml as a dependency

### LINKED ISSUE #260: PHP's XML extension required for Authentication
There's no dependency on ext-xml in composer.json although \Gitlab\HttpClient\Plugin\Authentication requires it for the global function utf8_encode() to be available. I'd suggest adding the dependency or using something else than utf8_encode().

GMTA:
Without ext-xml a fatal error can occur because of the absence of utf8_encode().

Fixes #260

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
 	],
 	"require": {
 		"php": "^5.6 || ^7.0",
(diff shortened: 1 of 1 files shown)


fbourigault:
Thank you!

PR #267, opened 2017-10-10
Clarify HTTPlug usage

curry684:
Previous wording suggested to always install the Guzzle6 adapter which rather defeats the purpose of messaging middleware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,15 @@ Installation
 Via [composer](https://getcomposer.org)
 
 ```bash
-composer require m4tthumphrey/php-gitlab-api php-http/guzzle6-adapter
+composer require m4tthumphrey/php-gitlab-api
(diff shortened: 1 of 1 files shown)


fbourigault:
At php-http, we are talking a lot about how to introduce HTTPlug in 3rd party libraries. We decided to put an opinionated choice with guzzle as the `composer require m4tthumphrey/php-gitlab-api` would fail with an error that would hurt developer experience.

So using `composer require m4tthumphrey/php-gitlab-api php-http/guzzle6-adapter` is fine as new users won't get an error and users aware of HTTPlug would know what to do here.

curry684:
Ok, clear, perhaps this is also something to take up with Composer - it would be able to give friendly errors based on suggestions perhaps.

PR #270, opened 2017-11-10
adding parameters on repositories->branches 

TwanoO67:
Example: usefull to set the per_page param
Otherwise this method only respond with a maximum of 20 branches


fbourigault:
Thank you!

TwanoO67:
Can we have a new release for that ?
(We need that on a "vinkla/gitlab" project, which follow your version as so: 
"m4tthumphrey/php-gitlab-api": "^9.0" )
Thank you :)

fbourigault:
I released 9.6.0 today which include those changes.

PR #274, opened 2017-11-17
Add support for Deployments and Environments api

zegenie:
Added basic support for fetching information about Deployments and Environments, since we needed this information in one of our deploy tools.

### DIFF
--- a/lib/Gitlab/Api/Deployments.php
+++ b/lib/Gitlab/Api/Deployments.php
@@ -2,22 +2,22 @@
 
 class Deployments extends AbstractApi
 {
-    /**
-     * @param int $project_id
-     * @return mixed
-     */
-    public function all($project_id)
-    {
-        return $this->get($this->getProjectPath($project_id, 'deployments'));
-    }
+	/**
+	 * @param int $project_id
+	 * @return mixed
+	 */
+	public function all($project_id)
+	{
+		return $this->get($this->getProjectPath($project_id, 'deployments'));
+	}
 
-    /**
-     * @param int $project_id
-     * @param string $deployment_id
-     * @return mixed
-     */
-    public function show($project_id, $deployment_id)
-    {
-        return $this->get($this->getProjectPath($project_id, 'deployments/'.$deployment_id));
-    }
+	/**
+	 * @param int $project_id
+	 * @param string $deployment_id
+	 * @return mixed
+	 */
+	public function show($project_id, $deployment_id)
+	{
+		return $this->get($this->getProjectPath($project_id, 'deployments/' . $deployment_id));
+	}
 }
--- a/lib/Gitlab/Api/Deployments.php
+++ b/lib/Gitlab/Api/Deployments.php
@@ -4,11 +4,13 @@ class Deployments extends AbstractApi
 {
     /**
      * @param int $project_id
+     * @param array $parameters
      * @return mixed
      */
-    public function all($project_id)
+    public function all($project_id, array $parameters = [])
     {
-        return $this->get($this->getProjectPath($project_id, 'deployments'));
+        $resolver = $this->createOptionsResolver();
+        return $this->get($this->getProjectPath($project_id, 'deployments'), $resolver->resolve($parameters));
     }
 
     /**
--- a/lib/Gitlab/Api/Deployments.php
+++ b/lib/Gitlab/Api/Deployments.php
@@ -2,22 +2,22 @@
 
 class Deployments extends AbstractApi
 {
-	/**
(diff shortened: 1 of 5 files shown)


lib/Gitlab/Api/Environments.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        return $this->get($this->getProjectPath($project_id, 'environments'));
+    }
+
+    /**
+     * @param int $project_id
+     * @param array $params
+     * @return mixed
+     */
+    public function create($project_id, array $params = array())
```
m1guelpf: Could you use OptionsResolver here?

zegenie: I was looking at other `create()` methods and I can't see any using the `OptionsResolver` - was just going with what seemed like the convention. Will have a look and update if possible.

zegenie:
Updated with using `OptionsResolver` for resolving `create()` parameters

lib/Gitlab/Api/Deployments.php
```diff
@@ -0,0 +1,23 @@
+<?php namespace Gitlab\Api;
+
+class Deployments extends AbstractApi
+{
+    /**
+     * @param int $project_id
+     * @return mixed
+     */
+    public function all($project_id)
```
fbourigault: Could use the `OptionsResolver` like in https://github.com/m4tthumphrey/php-gitlab-api/blob/master/lib/Gitlab/Api/Repositories.php#L13 to allow pagination?

lib/Gitlab/Api/Environments.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class Environments extends AbstractApi
+{
+    /**
+     * @param int $project_id
+     * @return mixed
+     */
+    public function all($project_id)
```
fbourigault: Same here for pagination.

lib/Gitlab/Api/Environments.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @param int $project_id
+     * @param array $parameters
+     * @return mixed
+     */
+    public function create($project_id, array $parameters = array())
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefined('name')
```
fbourigault: According to the documentation this parameter is required. Could you use `setRequired` instead?

lib/Gitlab/Api/Environments.php
```diff
@@ -0,0 +1,43 @@
...
+            ->setAllowedTypes('name', 'string');
+        $resolver->setDefined('slug')
```
fbourigault: I don't find any reference in documentation about this `slug` argument. Could you provide some official input that prove it exists? I used to reject parameters not officially documented as those may be painful to maintain in the future.

lib/Gitlab/Api/Environments.php
```diff
@@ -0,0 +1,43 @@
...
+     * @return mixed
+     */
+    public function all($project_id)
+    {
+        return $this->get($this->getProjectPath($project_id, 'environments'));
+    }
+
+    /**
+     * @param int $project_id
+     * @param array $parameters
```
fbourigault: Could you add details about this argument like in:
https://github.com/m4tthumphrey/php-gitlab-api/blob/35ecf9b82984beace5eaea0df4fcbe4f77e2c285/lib/Gitlab/Api/Repositories.php#L179-L194

lib/Gitlab/Api/Environments.php
zegenie: Done.

lib/Gitlab/Api/Deployments.php
zegenie: Done. I wasn't sure whether this had pagination so I just left it out. I also added `$parameter` in the docblock for `Repositories` used as example as it was missing.

lib/Gitlab/Api/Environments.php
zegenie: Done.

lib/Gitlab/Api/Environments.php
zegenie: Sure thing. Done, using github API description for the details.

lib/Gitlab/Api/Environments.php
zegenie: I removed this as I had just misunderstood the return value coming back from the api. It was never a valid parameter. Good catch.

fbourigault:
Thank you!

PR #276, opened 2017-11-21
Branches containing a slash break v4 branch endpoints. #263

### LINKED ISSUE #263: Branches containing a slash break v4 branch endpoints
The encoding of branch names in API v4 has changed from unencoded slashes to encoded as "%2f". As such, getting branch information for a branch like "env/production" is broken in the latest version.

Ironically this means #86 basically needs to be reverted.

Removing the `str_replace` call in [Repositories](https://github.com/m4tthumphrey/php-gitlab-api/blob/master/lib/Gitlab/Api/Repositories.php#L432) fixes the issue for me but I'm not sure if there are any other places where this needs to be done.

b0nd0:
Hello.
This is my attempt to fix #263.

I tested only `Delete branch` endpoint, cause this is the only one we are using here right now.
I completely removed `encodeBranch` function being guided by #86 pull request, which we kinda reverting now.

upd:
by "tested" i mean on real gitlab installation, not just unit.

### DIFF
--- a/lib/Gitlab/Api/Repositories.php
+++ b/lib/Gitlab/Api/Repositories.php
@@ -22,7 +22,7 @@ class Repositories extends AbstractApi
      */
     public function branch($project_id, $branch_id)
     {
-        return $this->get($this->getProjectPath($project_id, 'repository/branches/'.$this->encodeBranch($branch_id)));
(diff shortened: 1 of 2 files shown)


b0nd0:
any thoughts?

fbourigault:
Thank you!

PR #279, opened 2017-12-14
Suggest OAuth2-GitLab client

curry684:
Since many use cases of the API client involve acting on behalf of a consenting user the manual should recommend how to do that.


curry684:
@m1guelpf explanation?

PR #280, opened 2017-12-15
Pagination of project members

tobiasetter:
GitLab API per default defines 20 items per_page for the returned result. Now if a project has more than 20 members we have an issue with increasing the number of items to max i.e. 100.

Therefore I complemented your code.

Many thanks!


tobiasetter:
Sorry, I just tried to pull it using an old version.
I will create a new fork and do my changes again.

Thanks for your understanding =)

PR #285, opened 2018-01-04
add share functionality

Artistan:
## reference
https://docs.gitlab.com/ee/api/projects.html#share-project-with-group

### DIFF
--- a/lib/Gitlab/Api/Projects.php
+++ b/lib/Gitlab/Api/Projects.php
@@ -676,10 +676,10 @@ class Projects extends AbstractApi
         $resolver = $this->createOptionsResolver();
 
         $resolver->setDefined('group_id')
-            ->setAllowedTypes('group_id','int');
+            ->setAllowedTypes('group_id', 'int');
 
         $resolver->setDefined('group_access')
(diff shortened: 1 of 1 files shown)


lib/Gitlab/Api/Projects.php
```diff
@@ -665,4 +665,24 @@ public function deployment($project_id, $deployment_id)
...
     }
+    
+    /**
+     * @param mixed $project_id
+     * @param array $params
+     * @return mixed
+     */
+    public function addShare($project_id, array $params = array())
+    {
+        return $this->post($this->getProjectPath($project_id, 'share'), $params);
```
m1guelpf: Can you use our options resolver for the `$params` array?

Artistan:
@m1guelpf - updated with resolver.

m1guelpf:
@Artistan Could you also add support for the `expires_at` parameter?

Artistan:
@m1guelpf - updated with required and expires_at

It looks like a lot of the normalizers are broken if my test is any indication.
http://symfony.com/doc/current/components/options_resolver.html#option-normalization
a lot of the normalizers I saw do not have the "Option" as the first parameter

PR #281, opened 2017-12-15
Pagination of Project Members

tobiasetter:
GitLab API per default defines 20 items per_page for the returned result. Now if a project has more than 20 members we have an issue with increasing the number of items to max i.e. 100.

Therefore I complemented your code.

Many thanks!

PR #286, opened 2018-01-04
fix members request with empty query.

Artistan:
members request does not require query and is actually broken right now if you provide an empty query
https://gitlab.com/gitlab-com/support-forum/issues/2878

### DIFF
--- a/lib/Gitlab/Api/Projects.php
+++ b/lib/Gitlab/Api/Projects.php
@@ -246,7 +246,7 @@ class Projects extends AbstractApi
      */
     public function members($project_id, $username_query = null)
     {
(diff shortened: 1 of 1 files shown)


m1guelpf:
@fbourigault Wouldn't this be fixed using `OptionsResolver`?

Artistan:
@m1guelpf - how is this?  I think you still need to check if null?

fbourigault:
Could you add a test to ensure we are not sending a `null` query?

tobiasetter:
@fbourigault @m1guelpf @Artistan This pull request would also be resolved by my pull request: #281

PR #290, opened 2018-01-25
Model Group projects should return Projects

carsonreinke:
This looks like a typo


carsonreinke:
@m1guelpf This should be good, if not let me know.

PR #293, opened 2018-02-07
added epics api

cdaguerre:


m4tthumphrey:
Thanks!

What about tests and models?

lib/Gitlab/Api/Epics.php
```diff
@@ -0,0 +1,58 @@
...
+     * @param int $group_id
+     * @param array $parameters 
+     *
+     * @return mixed
+     */
+    public function all($group_id, array $parameters = [])
+    {
+        $resolver = $this->createOptionsResolver();
+
+        return $this->get('groups/'.$this->encodePath($group_id).'/-/epics', $resolver->resolve($parameters));
```
glensc, line 15: this is just wrong? there's no such API path like `/-/epics`:
- https://docs.gitlab.com/ee/api/epics.html

I'm sure OP thought web_url paths in gitlab.com web are api paths to be used here.

GrahamCampbell:
Closing due to inactivity. If anyone would still like this, please submit a new PR with tests.

Issue #294, opened 2018-02-09
visibility param for snippet creating is missing


mirkoschmidt:
https://github.com/m4tthumphrey/php-gitlab-api/pull/96
This pull request is important for creating snippets.

Edit: I have create a new pull request #296

PR #296, opened 2018-02-14
Add snippet visibility for creating

mirkoschmidt:
Add required param visibility for creation of snippets.

#294 close


GrahamCampbell:
This was a major breaking change.

glensc:
@GrahamCampbell because the new parameter is required (has no default value)?

PR #300, opened 2018-03-20
Switched to PHPUnit 5

GrahamCampbell:
Let's see if this passes. :)


m1guelpf:
@GrahamCampbell Thank you!

PR #310, opened 2018-04-12
omit query parameter on null value in projects api

### LINKED ISSUE #259: Project members result empty when username_query is not supplied
When I try to get the project's members from the API, without supplying a `username_query` (I want to get all members), the result from the client is empty. When I call the URL (`https://gitlab.com/api/v4/projects/[projectID]/members`) manually, I get all members. When I do supply a `username_query` I do get the users that comply to that query.

So something seems to be not working correctly.

Code without query:

```php
$members = $client->projects()->members($projectId);
// $members is now an empty array
```

Code with query:

```php
$members = $client->projects()->members($projectId, 'willem');
// $members now contains one member: me
```

dariuskt:
closes #259

### DIFF
--- a/lib/Gitlab/Api/Projects.php
+++ b/lib/Gitlab/Api/Projects.php
@@ -247,9 +247,12 @@ class Projects extends AbstractApi
      */
     public function members($project_id, $username_query = null)
     {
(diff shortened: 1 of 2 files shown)


tobiasetter:
@dariuskt  This pull request would also be resolved by my pull request: #281

PR #314, opened 2018-04-24
Update the Version of API to work with v3 & v4

tom-sapletta-com:
Before was not worked with Version API v3


curry684:
This is intentionl: https://github.com/m4tthumphrey/php-gitlab-api#versioning

v3 and v4 are not compatible so not possible to support them in the same release.

PR #317, opened 2018-04-26
removed pagination parameters default values and validation for project members

tobiasetter:
After the merge of #281 I realized that we don't need the validation of the parameters "page" and "per_page" since it will be validated by the gitlab api itself.

Therefore I removed the validation and default values for these parameters but also added a test for getting the project members with pagination.

@m1guelpf  Sorry that I've noticed only now.


m1guelpf:
Thank you for your fix, @tobiasetter!

PR #318, opened 2018-04-26
encode $job_name

lowtower:
In the method `artifactsByRefName()` the $job_name is not encoded.
I guess that should be done to avoid errors


m1guelpf:
Thank you for your contribution, @lowtower!

Issue #326, opened 2018-05-21
fix build query


stas17:
we get `500 Internal Server Error`
if not set not required params 

for example
```

    /**
     * @param int $project_id
     * @param string $source
     * @param string $target
     * @param string $title
     * @param int $assignee
     * @param int $target_project_id
     * @param string $description
     * @return mixed
     */
    public function create($project_id, $source, $target, $title, $assignee = null, $target_project_id = null, $description = null)
    {
        return $this->post($this->getProjectPath($project_id, 'merge_requests'), array(
            'source_branch' => $source,
            'target_branch' => $target,
            'title' => $title,
            'assignee_id' => $assignee,
            'target_project_id' => $target_project_id,
            'description' => $description
        ));
    }
```
this function are building this `source_branch=foo&target_branch=barr&title=for_example&assignee_id=&target_project_id=&description='`
query without nullable fields

PR #327, opened 2018-05-21
fix build query

stas17:
#326


20uf:
Question for @m4tthumphrey , why not simplify using [http_build_query ](http://php.net/manual/fr/function.http-build-query.php) ?

PR #336, opened 2018-07-04
add a method to pull refs (tags and branches) for a given commit

mlhess:
This allows us to get the refs for a given commit.


mlhess:
The test would be 100% faked as we are not pulling any real data nor does the data persist unless I am missing something.

m1guelpf:
Yeah, but we need tests to ensure, for example, we don't remove a parameter and break something accidently.

drumm:
PR #343 adds an additional test for this.

PR #346, opened 2018-08-26
Add methods for managing emails for users

drumm:
This adds integrations for:
- https://docs.gitlab.com/ce/api/users.html#list-emails-for-user
- https://docs.gitlab.com/ce/api/users.html#add-email-for-user
- https://docs.gitlab.com/ce/api/users.html#delete-email-for-given-user


drumm:
This resolves #106

PR #349, opened 2018-09-07
Update parameters according to https://docs.gitlab.com/ee/api/commits…

marcoboers94:
Allow the "all" and "with_stats" parameters for the commits API endpoint.
As documented here: https://docs.gitlab.com/ee/api/commits.html


m1guelpf:
Update tests please

PR #357, opened 2018-09-26
Get the original array of model

marcteyssier:
I add an original data variable to the full data set in the hydrater.
With the function getOriginalData, it's easy to retrieve the real data and get an added parameter.


GrahamCampbell:
Closing due to inactivity.

PR #367, opened 2018-10-22
Properly escape $tag_name in Tags API

drumm:
Tag names might contain slashes, for example.


m1guelpf:
@drumm Tests are failing

drumm:
It looks like the code had been testing →create() 3 times instead of →createRelease() & →updateRelease(). I pushed a commit to update that, which Travis doesn't seem to have picked up.

PR #369, opened 2018-10-23
Implement project transfer API

drumm:
This implements another project transfer API: https://docs.gitlab.com/ce/api/projects.html#transfer-a-project-to-a-new-namespace

This was added by https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/20122, seems like permissions might be different from the other ways to transfer a project.

(For our instance of GitLab, this API returns normally. While `…->api('groups')->transfer(…)` does transfer the project, but the HTTP request times out.)

### DIFF
--- a/lib/Gitlab/Api/Projects.php
+++ b/lib/Gitlab/Api/Projects.php
@@ -398,6 +398,15 @@ class Projects extends AbstractApi
 
     /**
      * @param int $project_id
+     * @param mixed $namespace
+     * @return mixed
+     */
+    public function transfer($project_id, $namespace) {
+        return $this->put($this->getProjectPath($project_id, 'transfer'), ['namespace' => $namespace]);
+    }
+
+    /**
+     * @param int $project_id
      * @return mixed
      */
     public function deployKeys($project_id)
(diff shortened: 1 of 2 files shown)


test/Gitlab/Tests/Api/TagsTest.php
```diff
@@ -94,10 +94,10 @@ public function shouldCreateRelease($releaseName, $description, $expectedResult)
         $api = $this->getApiMock();
         $api->expects($this->once())
             ->method('post')
-            ->with('projects/1/repository/tags/' . urlencode($releaseName) . '/release', $params)
+            ->with('projects/1/repository/tags/' . str_replace(['/', '.'], ['%2F', '%2E'], $releaseName) . '/release', $params)
```
m1guelpf: why this change?

drumm: This was added by https://github.com/m4tthumphrey/php-gitlab-api/pull/368, rebasing so this pull request does not have extra changes.

`AbstractApi::encodePath()` explicitly encodes dots, this lines up with that for testing. https://github.com/m4tthumphrey/php-gitlab-api/blob/52965f4af56bb39dcb5393606b612424eab2a176/lib/Gitlab/Api/AbstractApi.php#L170

PR #376, opened 2018-11-08
Tag message and release properties

bmalex88:
These changes are for https://github.com/m4tthumphrey/php-gitlab-api/issues/375.
Please, accept them if everything allright, so we can get release notes and tags mesages. Thanks


bmalex88:
Is this enough?

PR #381, opened 2018-11-27
[9.18] Allow to search and find issues by "assignee_id"

koehnlein:
`assignee_id` is a valid parameter in GitLab's Issue API https://docs.gitlab.com/ce/api/issues.html#list-issues

It would be great, if you would allow your library to use it as filter parameter.


m1guelpf:
Please add tests

GrahamCampbell:
Thanks for this. It looks like this was already partially implemented. I have rebased and merged your addition to the phpdoc.

PR #388, opened 2019-01-01
Test on php 7.3 too

GrahamCampbell:


m1guelpf:
Thanks @GrahamCampbell!

PR #392, opened 2019-01-14
Support user avatar file upload

drumm:
This adds support for uploading a user's avatar via a new files argument. It works like `…->update(…, array('avatar' => 'path/to/avatar'))`

The code in `AbstractApi::put()` is copied from `AbstractApi::post()`.

I wasn’t able to find any existing testing of file uploads. Please let me know if there is a good way to add tests.


drumm:
I wasn’t able to find any existing testing of file uploads. Please let me know if there is a good way to add tests.

m1guelpf:
@drumm You can probably copy an existing test and change the endpoint and parameters

drumm:
I have added a test to cover this.

PR #401, opened 2019-02-16
Added assignee_id to defined options

brandonferens:
This added option allows only issues assigned to a specific person to be returned.


m1guelpf:
Please add a test for this

brandonferens:
@m1guelpf I've added a test for the `asssignee_id` parameter.

PR #411, opened 2019-03-11
add Badge model missing properties

sluc23:


sluc23:
well test were actually correct, using all Badge parameters, I've just missed them in the model class

https://github.com/m4tthumphrey/php-gitlab-api/blob/master/test/Gitlab/Tests/Api/ProjectsTest.php#L1606

PR #416, opened 2019-03-27
Allow setting path and name when forking a Project

ijaureguialzo:
GitLab API allows two optional params when forking, path and name: https://docs.gitlab.com/ee/api/projects.html#fork-project

Added them to the validator so, when forking a repo, it can be renamed and placed in a different path.


ijaureguialzo:
Tests added.

PR #418, opened 2019-04-12
List all members of a group or project including inherited members

otrelin:


otrelin:
@m1guelpf 
Hi, Miguel!
I just saw this commit https://github.com/m4tthumphrey/php-gitlab-api/commit/553fab6c66a8b93731d900b37076bc1b9d0e54ec
It's the same as I've done. But I've done also for groups. Apparently, it is worth leaving only one method for projects and it should be the same style as for groups.
Please, decide which method to leave and make a release. Because different people write the same functional.

m1guelpf:
Tagged

PR #426, opened 2019-04-29
Add a SHA property/filter for pipelines

laurenzgamper:
I need to search the pipelines for a sha of a specific commit. This change adds a parameter to filter the pipelines for a sha of a commit.

If you have questions or require changes, please let me know.


laurenzgamper:
I added a test like the shouldGetPipelinesWithBooleanParam, though I'm not sure how it tests anything...

PR #431, opened 2019-05-04
Fixed redirects for downloading artifacts

Wouter0100:
When downloading artifacts, I'm being redirected to an object store at Google. At this point the ApiVersion plugin currently adds the API version to Google's URL, which causes errors.

By keeping track of the last request, we know if it is being redirected. The following request will overwrite the last stored state. To be able to keep track of redirects, we need to be after the RedirectPlugin.


Wouter0100:
Changes pushed.

m1guelpf:
Thank you for your contribution, @Wouter0100!

Issue #446, opened 2019-10-02
Model\Label() id is not accessible but it is needed for board list creation


DBLaci:
I can list the labels for a project using $project->labels() where $project is a Model\Project object. But the label object contains only the name and the color of the label, not the id.

I need the label id however for board list creation.

I want to maintain labels and board lists for projects with a script, hence I need this.

m1guelpf:
Feel free to open a PR that fixes this

PR #447, opened 2019-10-04
#446 - label model id

DBLaci:
https://github.com/m4tthumphrey/php-gitlab-api/issues/446


DBLaci:
Can you please check again? I updated the tests I think are affected (was not breaking tests tough), and created a Label model test with the very basic tests.

PR #457, opened 2019-11-09
Added sudo method for method chaining.

leoleoasd:
Usage:
```
$client  = \Gitlab\Client::create('http://git.yourdomain.com')
    ->authenticate('your_gitlab_token_here', \Gitlab\Client::AUTH_URL_TOKEN);
$client->foo(); // normal requests;
$client->sudo("username")->bar(); // sudo requests;
$client->foo(); //normal requests;
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
 		"php-http/discovery": "^1.2",
 		"php-http/httplug": "^1.1 || ^2.0",
 		"php-http/multipart-stream-builder": "^1.0",
-		"symfony/options-resolver": "^2.6 || ^3.0 || ^4.0"
+		"symfony/options-resolver": "^2.6 || ^3.0 || ^4.0",
+		"myclabs/deep-copy": "^1.9"
 	},
 	"require-dev": {
 		"guzzlehttp/psr7": "^1.2",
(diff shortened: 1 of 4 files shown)


m1guelpf:
Can we add some tests for this?

leoleoasd:
> Can we add some tests for this?
> @m1guelpf 

I am a freshman at university and I don’t know exactly how phpunit works.  TwT

I wrote this feature to fit my own requirements ( something like `GitLab::sudo(1)->foo()` in Laravel), i think it may be useful for some others so i made this pull request.

I’ll read some tutorials to see if i can add some tests, but i cant guarantee i can figure it out.

I haven’t tested this code yet, and my project using this feature is still under development, but i will test it manually in my development environment this weekend.

Also, sorry for my English. (I am a Chinese student)

leoleoasd:
Tested manually on my personal development environment.
I'm still trying to figure out how to write a unit test for this feature.

GrahamCampbell:
This doesn't look right. We certainly don't want to copy everything, like the http client.

GrahamCampbell:
👎 for merging this. The usage I'd recommend would be to build a new object with sudo enabled, from the start.

glensc:
this is absolutely not needed, looking at the implementation:

https://github.com/m4tthumphrey/php-gitlab-api/blob/9.17.0/lib/Gitlab/Client.php#L420-L434


you can just replace `sudo` state by re-authenticating:

```php
$sudo = true;
$client->authenticate('your_gitlab_token_here', \Gitlab\Client::AUTH_URL_TOKEN, $sudo);

$sudo = false;
$client->authenticate('your_gitlab_token_here', \Gitlab\Client::AUTH_URL_TOKEN, $sudo);

```

PR #468, opened 2019-12-19
implemented Environments::show()

### LINKED ISSUE #467: Unable to get last deployment of an environment
There is no `show()` method implemented in `API\Environments`. 
This would provide access to the additional keys in a single environment fetch.

webmaster777:
closes #467


webmaster777:
This has also been tested against our own gitlab instance 

image

webmaster777:
Travis fails, but this seems not related to this change.

PR #476, opened 2020-01-16
[9.18] Adding 'stats' property to Commit model

AtCliffUnderline:
So, when I'm trying to get list of commits with 'with_stats' property - the array 'stats' which Gitlab returns dissapears because no such property added.


m1guelpf:
@AtCliffUnderline Can you add tests for this?

GrahamCampbell:
Thanks for this. :)

GrahamCampbell:
I've added the relevant property to the class level phpdoc too.

PR #490, opened 2020-02-28
[9.18] Adding a single group membership endpoint

dzink:
See https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project

This only implements this function for groups.

### DIFF
--- a/lib/Gitlab/Api/Groups.php
+++ b/lib/Gitlab/Api/Groups.php
@@ -120,6 +120,20 @@ class Groups extends AbstractApi
     }
 
     /**
+     * @param int   $group_id
+     * @param int   $user_id
+     *
+     * @return mixed
+     */
+    public function member($group_id, $user_id)
+    {
+        $resolver = $this->createOptionsResolver();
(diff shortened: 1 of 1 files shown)


lib/Gitlab/Api/Groups.php
```diff
@@ -119,6 +119,20 @@ public function members($id, array $parameters = [])
...
+     * @param int   $group_id
+     * @param int   $user_id
+     *
+     * @return mixed
+     */
+    public function member($group_id, $user_id)
+    {
+        $resolver = $this->createOptionsResolver();
+        $resolver->setDefined('query');
+
```
glensc, line 132: $resolver is unused, too much copy-pasting and not testing your added functionality?

I personally would add @see annotation and link to api doc, so could verify what parameters are supported

GrahamCampbell:
Thanks for starting work on this @dzink. Are you able to finish off this PR please?

GrahamCampbell:
Thanks for this. I've finished this off now. Note that we've already implemented this on the projects API.

PR #492, opened 2020-03-06
List all members of a group or project including inherited members with optional user id filter

SirFlip:


GrahamCampbell:
This introduced a major breaking change to the method signature. I will have to revert this PR.

test/Gitlab/Tests/Api/GroupsTest.php
```diff
@@ -199,13 +199,29 @@ public function shouldGetAllMembers()
         $api = $this->getApiMock();
         $api->expects($this->once())
             ->method('get')
-            ->with('groups/1/members/all')
+            ->with('groups/1/members/all/')
```
GrahamCampbell, line 202: Trailing slashes matter! This is not the same URI.

lib/Gitlab/Model/Project.php
```diff
@@ -217,6 +217,30 @@ public function remove()
...
 
+    /**
+     * @param integer|null $user_id
+     * @param bool $all
+     * @return array|User
+     */
+    public function allMembers($user_id = null, $all = false)
+    {
+        if ($all) {
+            $data = (new \Gitlab\ResultPager($this->client))->fetchAll($this->client->projects(), "allMembers", [$this->id, $user_id]);
```
GrahamCampbell, line 228: We don't use the result pager in models due to separation of concerns.

PR #498, opened 2020-04-27
Add optional parameter first_parent to the commits endpoint

dhzavann:


glensc:
can you check whether the other PR does the same:
- https://github.com/m4tthumphrey/php-gitlab-api/pull/502

I'm wondering does the boolean normalized break something or not?

dhzavann:
@glensc Hi, I was thinking of adding the normalization more or less just like that, but decided not to just because the other parameters in the function didn't use it for some reason and I didn't want to spend the time digging into why.

My guess is that it was just an oversight when those were added and that your PR is better.

glensc:
@dhzavann can you confirm the boolean normalization is valid (in my PR) and close your PR?

GrahamCampbell:
Thanks for your work on this. I think this is superseded by https://github.com/m4tthumphrey/php-gitlab-api/pull/502.

PR #499, opened 2020-04-28
[11.1] Missing parameters from issues endpoint and missing memberships endpoint

karsa-mistmere:
I've added missing parameters to Issues (https://docs.gitlab.com/ee/api/issues.html)
I've also added a memberships method (https://docs.gitlab.com/ee/api/users.html#user-memberships-admin-only)

### DIFF
--- a/lib/Gitlab/Api/Issues.php
+++ b/lib/Gitlab/Api/Issues.php
@@ -1,5 +1,7 @@
 <?php namespace Gitlab\Api;
 
+use Symfony\Component\OptionsResolver\Options;
+
 class Issues extends AbstractApi
 {
     /**
@@ -328,7 +330,7 @@ class Issues extends AbstractApi
         $resolver = parent::createOptionsResolver();
 
         $resolver->setDefined('state')
-            ->setAllowedValues('state', ['opened', 'closed'])
+            ->setAllowedValues('state', ['all', 'opened', 'closed'])
         ;
         $resolver->setDefined('labels');
         $resolver->setDefined('milestone');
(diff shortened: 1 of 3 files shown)


lib/Gitlab/Api/Issues.php
```diff
@@ -352,6 +354,26 @@ protected function createOptionsResolver()
             ->setAllowedTypes('assignee_id', 'integer')
         ;
 
+        $datetimeNormalizer = function (Options $resolver, \DateTimeInterface $value) {
+            return $value->format('c');
+        };
+
+        $resolver->setDefined('created_after')
+            ->setAllowedTypes('created_after', \DateTimeInterface::class)
```
GrahamCampbell, line 362: Please update the phpdoc of all associated methods so `created_after` is described as `string|\DateTimeInterface`, and similarly for the other cases.

lib/Gitlab/Api/Users.php
```diff
@@ -140,6 +140,23 @@ public function usersProjects($id, array $parameters = [])
...
     }
 
+    /**
+     * @param int $id
+     * @param array $parameters {
+     *     @var string $type                        Filter memberships by type. Can be either Project or Namespace
+     * }
+     * @return mixed
+     */
+    public function memberships($id, array $parameters = [])
```
GrahamCampbell, line 150: Please could you send this change in a separate PR.

lib/Gitlab/Api/Issues.php
GrahamCampbell: I think there may be a few other places in the codebase that also need addressing.

GrahamCampbell:
Ping @karsa-mistmere.

GrahamCampbell:
Friendly ping @karsa-mistmere.

PR #501, opened 2020-04-30
[9.18] Add method to access repository/commits/.../discussions

glensc:
Gets a list of all discussion items for a single commit.

See https://docs.gitlab.com/ee/api/commits.html#get-a-single-commit

not to be confused with discussions API, which is different:
- https://docs.gitlab.com/ee/api/discussions.html#list-project-commit-discussion-items

Example:
- https://gitlab.com/gitlab-org/gitlab/-/commit/695c29abcf7dc2eabde8d59869abcea0923ce8fa#note_334686748
- https://gitlab.com/api/v4/projects/gitlab-org%2Fgitlab/repository/commits/695c29abcf7dc2eabde8d59869abcea0923ce8fa/discussions

needed for:
- https://github.com/glensc/slack-unfurl-gitlab/pull/10

so just adds single method getRepositoryCommitDiscussions to projects api.


glensc:
there's also this endpoint:
- https://gitlab.com/api/v4/projects/gitlab-org%2Fgitlab/repository/commits/695c29abcf7dc2eabde8d59869abcea0923ce8fa/comments

```php
$comments = $client->repositories->commitComments($project_id, $commit);
```

but that won't work for me, as I need to lookup comment by id, and the note id is not available on that endpoint

GrahamCampbell:
Thanks for the PR @glensc. Please add a test, and I'll merge this. :)

glensc:
@GrahamCampbell what do you want to test for this code?:

```php
    public function getRepositoryCommitDiscussions($project_id, $commit_id)
    {
        return $this->get($this->getProjectPath($project_id, 'repository/commits/' . $this->encodePath($commit_id)) . '/discussions');
    }
```

I'm reluctant to add tests that just test that arguments I force feed come out the same. it's basically testing only that URL builder works as expected and that should be covered with just adding tests to 
`getProjectPath` method.

see (also) my rant at:
- https://github.com/m4tthumphrey/php-gitlab-api/issues/481

GrahamCampbell:
I'm aware, however bugs do actually come up during refactoring where even those sorts of tests are useful.

GrahamCampbell:
Thanks. :)

glensc:
Sorry, that didn't get to adding the tests part :)

glensc:
waiting for the release, so can remove the monkey patch:
- https://github.com/glensc/slack-unfurl-gitlab/blob/8578dfca90a79beca9e98c410691f1b98cc9bd7c/src/Route/Note.php#L118-L138

GrahamCampbell:
I'd appreciate if you could test with `9.18.*@dev` before I release, in case there are any bugs introduced since 9.17.0 (or indeed in 9.17.1). I did a fair bit of refactoring and marking code as deprecated.

PR #551, opened 2020-07-06
[9.18] Update repo URL and switch to GitHub Actions

GrahamCampbell:


.gitignore
```diff
@@ -1,5 +1,4 @@
-/composer.lock
-/phpstan.neon
-/phpstan.phar
-/phpunit.xml
-/vendor
+composer.lock
+phpstan.neon
+phpunit.xml
+vendor
```
glensc, line 4: why going back to relative paths? if I have mistakenly created composer.lock in non-root dir, I would `git status` to show me that.

GrahamCampbell: Because these directories really do appear in places other than the root directory now, such as within the `vendor-bin` directory.

glensc: `/vendor-bin/*/vendor/` ... :)

altho, in my project I'll try just placing under vendor:
- https://github.com/bamarni/composer-bin-plugin#change-directory

```
"target-directory": "vendor/vendor-bin"
```

GrahamCampbell: But I need the composer.json files that are in the vendor-bin directory.

GrahamCampbell: Meh, I think what we have is OK. If there are any genuine issues caused, we can certainly revisit this.

PR #557, opened 2020-07-16
[9.19] Added user events

jvandervalk:
I've added a missing API call to /users/:id/events to retrieve events for a specified user
See https://docs.gitlab.com/ee/api/events.html#get-user-contribution-events

### DIFF
--- a/lib/Gitlab/Api/Users.php
+++ b/lib/Gitlab/Api/Users.php
@@ -486,5 +486,6 @@ class Users extends AbstractApi
             ->setAllowedValues('sort', ['asc', 'desc'])
         ;
 
-        return $this->get('users/'.$this->encodePath($user_id).'/events', $resolver->resolve($parameters));    }
+        return $this->get('users/'.$this->encodePath($user_id).'/events', $resolver->resolve($parameters));
+    }
 }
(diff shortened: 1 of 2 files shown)


lib/Gitlab/Api/Users.php
```diff
@@ -448,4 +448,44 @@ public function removeImpersonationToken($user_id, $impersonation_token_id)
     {
         return $this->delete('users/'.$this->encodePath($user_id).'/impersonation_tokens/'.$this->encodePath($impersonation_token_id));
     }
+
+    /**
+     * @param int|string $user_id
```
GrahamCampbell: when is the user id a string?

GrahamCampbell (commented):
Thanks for this. I just have one minor comment. :)

GrahamCampbell:
Excellent, thanks. 🍻

You can use this code by installing `m4tthumphrey/php-gitlab-api:^9.19@dev` until v9.19.0 is tagged. The current plan is to release v9.19.0 and v10.0.0 simultaneously in August.

lib/Gitlab/Api/Users.php
jvandervalk: Thought maybe if someone saved it in his own database, but now I have a closer look at User.php I don't see it elsewhere, I'll change it!

PR #563, opened 2020-08-07
Add method impersonate() to Client to act on a users behalf

spawnia:
This method conveniently takes care of both creating and revoking the impersonation token and allows making API calls while authenticated as the user.

If this idea is something you want to pursue further, I am happy to add tests and documentation.

### DIFF
--- a/lib/Gitlab/HttpClient/Builder.php
+++ b/lib/Gitlab/HttpClient/Builder.php
@@ -168,16 +168,20 @@ class Builder
      *
      * @param string $fqcn
      *
-     * @return void
+     * @return Plugin|null
      */
     public function removePlugin($fqcn)
     {
+        $plugin = null;
+
         foreach ($this->plugins as $idx => $plugin) {
             if ($plugin instanceof $fqcn) {
                 unset($this->plugins[$idx]);
                 $this->pluginClient = null;
             }
         }
+
+        return $plugin;
     }
 
     /**
--- a/lib/Gitlab/Client.php
+++ b/lib/Gitlab/Client.php
@@ -497,6 +497,12 @@ class Client
         $tokenResponse = $usersApi->createImpersonationToken($userId, 'gitlab-php-client', ['api']);
 
         $httpClientBuilder = $this->getHttpClientBuilder();
+
+        /**
+         * Can not be null, since we were able to authenticate to get the impersonation token.
+         *
+         * @var Authentication $previousAuthentication
+         */
         $previousAuthentication = $httpClientBuilder->removePlugin(Authentication::class);
 
         try {
--- a/lib/Gitlab/Client.php
+++ b/lib/Gitlab/Client.php
@@ -490,7 +490,7 @@ class Client
      *
      * @return $this
      */
-    public function impersonate(int $userId, callable $act)
+    public function impersonate($userId, callable $act)
     {
         $usersApi = $this->users();
 
--- a/lib/Gitlab/Client.php
+++ b/lib/Gitlab/Client.php
@@ -470,8 +470,43 @@ class Client
             @trigger_error(sprintf('Passing an invalid authentication method is deprecated since version 9.1 and will be banned in version 10.0.'), E_USER_DEPRECATED);
         }
 
-        $this->getHttpClientBuilder()->removePlugin(Authentication::class);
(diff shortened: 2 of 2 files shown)


lib/Gitlab/HttpClient/Builder.php
```diff
@@ -168,16 +168,20 @@ public function addPlugin(Plugin $plugin)
...
      * @param string $fqcn
      *
-     * @return void
+     * @return Plugin|null
      */
     public function removePlugin($fqcn)
     {
+        $plugin = null;
+
         foreach ($this->plugins as $idx => $plugin) {
```
spawnia, line 177: How about we store plugins as a map from FQCN to instance instead?

Not sure if adding a plugin twice is ever needed. If so, I can change this method to return an array of removed plugins instead.

GrahamCampbell:
9.19 please, not 9.18.

GrahamCampbell:
Why not just make impersonation a special type of authentication?

spawnia:
> Why not just make impersonation a special type of authentication?

Not sure if I understand what you mean correctly, how would you imagine that too work? In order to impersonate another user, you already have to be authenticated as an admin user.

Let me outline our use case, it might make the motivation and design decisions more clear:

We built automation to always open a merge request right away whenever someone pushes a branch. This works by sending a webhook on push events to our automation service.

As of now, the automation service authenticates with a static token that belongs to an admin user we call *gitlab-bot*. Unfortunately, the merge requests it opens do then also belong to that user. Subsequentially, the git history is borked, losing information about who initially created the commits.

What I want to achieve is that the user who pushed the branch is also the one who will create the merge request. Since we do not want to store personal access tokens for all users in our service, my idea was to use impersonation. Using the user id from the push event, we can get a temporary token to create the merge request, then discard it right after.

Doing this manually, the code looks like this:

```php
    protected function createMergeRequest(Push $push, string $source, string $target): void
    {
        $userId = $push->get('user_id');
        $userApi = $this->gitlab->users();

        $tokenResponse = $userApi->createImpersonationToken(
            $userId,
            self::IMPERSONATION_TOKEN_NAME,
            self::IMPERSONATION_TOKEN_SCOPES
        );

        try {
            $this->gitlab->authenticate($tokenResponse['token']);
            $this->gitlab
                ->mergeRequests()
                ->create(...);
        } finally {
            $this->gitlab->authenticate(env('GITLAB_BOT_TOKEN'));
            $userApi->removeImpersonationToken($userId, $tokenResponse['id']);
        }
    }
```

With the addition of this API, we will be able to simplify to:

```php
    protected function createMergeRequest(Push $push, string $source, string $target): void
    {
        $userId = $push->get('user_id');
        $this->gitlab->impersonate($userId, function (GitLabManager $gitlab) use ($push, $userId, $source, $target) {
            $gitlab
                ->mergeRequests()
                ->create(...);
        });
    }
```

The tricky bit of ensuring that the token is removed, even on failure, will be handled correctly every time.

glensc:
looking forward to this, @GrahamCampbell can you respond to @spawnia?

glensc:
Some other pattern that could be useful:
1. $impersonated = $gitlab->impersonate($userId) -> does clone of $gitlab and enables sudo tokem
2. $impersonated->__destruct() handles the token removal

this way there is no funny global state that needs to remove (on the original $gitlab object), you get a clone of the $gitlab with different credentials that you can use in parallel at the same time.

maybe this is more transparent to use than the callback wrapping

spawnia:
@glensc I have not thought of using `__destruct()`, looks like it could work. However, I do feel that this is approach is more magical. As of now, the semantics look simpler to me: The impersonation lasts during the callback and ends right after.

Cloning is quite messy in PHP (and probably in most languages that have mutable values). We can not guarantee a deep clone, since there can be user defined plugins.

GrahamCampbell:
I think the simplest approach is to just make this part of authentication, and don't allow for adhoc impersonation.

spawnia:
> I think the simplest approach is to just make this part of authentication, and don't allow for adhoc impersonation.

I don't know how to make this any simpler. Can you sketch out what solution you would like?

GrahamCampbell:
I don't think this is something I want to include in the package core, since it can be easily implemented with a helper in your app.

PR #578, opened 2020-09-25
[11.0] Add label id parameter to updateLabel method

minorOffense:
Fix updateLabel to require label id
Fix update  URL to include label id
See https://docs.gitlab.com/ee/api/group_labels.html#update-a-group-label

### DIFF
--- a/src/Api/Groups.php
+++ b/src/Api/Groups.php
@@ -315,9 +315,9 @@ class Groups extends AbstractApi
      *
      * @return mixed
      */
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks for the PR. Could you please rebase this and update the phpdoc too.

GrahamCampbell:
Hmm, this is actually a breaking change. Older versions of GitLab take the param via query params:

image

GrahamCampbell:
Applied a more comprehensive version of this to the 11.0 branch.

PR #579, opened 2020-09-25
fix(group): update label parameters

minorOffense:
Add label id
See https://docs.gitlab.com/ee/api/group_labels.html#update-a-group-label


GrahamCampbell:
Thanks for this. I will just merge this into 10.1.

PR #588, opened 2020-10-20
[9.19] Add with_merge_status_recheck option for fetching merge requests

drumm:
GitLab added `with_merge_status_recheck` in 13.0: https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests

Issue #573, opened 2020-09-09
[10.x] iids is invalid


andrey-helldar:
I am sending a request to get a list of comments for MR

Version 9.19 generated a request like this ([working](https://github.com/GitLabPHP/Client/blob/9.19/lib/Gitlab/HttpClient/Util/QueryStringBuilder.php#L26-L28)):
```
/api/v4/projects/31/merge_requests?iids[]=21
```

Version 10 makes a request like this ([doesn't work](https://github.com/GitLabPHP/Client/blob/10.0/src/HttpClient/Util/QueryStringBuilder.php#L25)):
```
/api/v4/projects/31/merge_requests?iids%5B0%5D=21
```

Therefore, the gitlab returns an error:
```
iids is invalid
```

```
{
    "error": {"type": "ValidationFailedException", "data": "iids is invalid"}, "info": {
    "exception": "Gitlab\\Exception\\ValidationFailedException",
    "file": "\/var\/www\/tunnel\/vendor\/m4tthumphrey\/php-gitlab-api\/src\/HttpClient\/Plugin\/GitlabExceptionThrower.php",
    "line": 60
}
```

GrahamCampbell:
Fixed.

andrey-helldar:
Thank you!

GrahamCampbell:
9.19.1, 10.0.1 and 10.1.0 will be tagged soon.

PR #589, opened 2020-10-21
[10.0] Add test to ensure iids does not become iids[0] in query string

drumm:
This is a test for #573 

I observed the same behavior when working around #587 / #588. With our GitLab 13.4.4 instance, I didn’t get an “iids is invalid” error; instead, the parameter was ignored and I got all merge requests.

### DIFF
--- a/tests/Api/MergeRequestsTest.php
+++ b/tests/Api/MergeRequestsTest.php
@@ -519,7 +519,7 @@ class MergeRequestsTest extends TestCase
         $api = $this->getApiMock();
         $api->expects($this->once())
             ->method('get')
-            ->with('projects/1/merge_requests', ['iids' => [2]])
+            ->with('projects/1/merge_requests', ['iids[]' => 2])
             ->will($this->returnValue($expectedArray))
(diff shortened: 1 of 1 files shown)


drumm:
Looks like this might need to test `QueryStringBuilder::build()` directly.

I don't see an option for `http_build_query()` to do this, so I don't know the best solution. There is a `preg_replace()` example in the comments at https://www.php.net/manual/en/function.http-build-query.php, it isn't very clean, but it doesn't look like the worst solution.

GrahamCampbell:
Right, so it looks like GitLab (or the ruby library they are using) is not following the RFCs properly, and is unable to parse URI formatted query strings properly, so we need to work around this.

GrahamCampbell:
I guess the solution is to revert 10.0 to use the builder from 9.19: https://github.com/GitLabPHP/Client/blob/9.19/lib/Gitlab/HttpClient/Util/QueryStringBuilder.php.

GrahamCampbell:
Done.

PR #591, opened 2020-10-26
Fix pager

Great-Antique:
Related issue: #592


GrahamCampbell:
Fixed in 10.0.2 and 10.1.1.

PR #596, opened 2020-10-30
[10.2] Add $variable_type to addVariable and updateVariable

### LINKED ISSUE #595: variable_type parameter missing in Projects::addVariable
It is currently not possible to add a project variable with the _variable type_ of _file_.
There would need to be an additional parameter in the Projects class https://github.com/GitLabPHP/Client/blob/0c7de45272e97da854162bcf85414cc047baa889/src/Api/Projects.php#L859

See: https://docs.gitlab.com/ee/api/project_level_variables.html#create-variable

Same goes for the update method on this line https://github.com/GitLabPHP/Client/blob/0c7de45272e97da854162bcf85414cc047baa889/src/Api/Projects.php#L886

See: https://docs.gitlab.com/ee/api/project_level_variables.html#update-variable

pselge-daparto:
Closes: #595

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -848,11 +848,11 @@ class Projects extends AbstractApi
     }
 
     /**
-     * @param int|string $project_id
-     * @param string $key
-     * @param string $value
-     * @param bool|null $protected
-     * @param string|null $environment_scope
+     * @param int|string          $project_id
+     * @param string              $key
+     * @param string              $value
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks. In future, don't worry about StyleCI. As per the contribution docs, StyleCI applies its fixed automatically when the PR is merged. ;)

GrahamCampbell:
10.2.0 has just been released.

PR #605, opened 2020-12-03
[10.4] Add min_access_level option to group search

icedream:
See https://docs.gitlab.com/ee/api/groups.html#list-groups.

The groups listing API endpoint allows specifying the `min_access_level` just the same way as with projects and users.


GrahamCampbell:
Great, thanks.

PR #607, opened 2020-12-11
[10.4] Update Users remove method to add params array to support hard_delete

jeffersonmartin:
This adds the `$params` array to `Api/Users` `remove()` to support optional parameters.

In GitLab v9.3, `hard_delete` was added as an optional parameter to the `DELETE /users/:id` API endpoint.

* [GitLab API Documentation](https://docs.gitlab.com/ee/api/users.html#user-deletion)
* [GitLab API v9.3 Commit](https://gitlab.com/gitlab-org/gitlab/-/commit/c890c6aaf2939bc19292947bd8268d724fa7ddce)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [10.4.0]: https://github.com/GitLabPHP/Client/compare/10.3.1...10.4.0
 
 * Add min_access_level option to group search
+* Added params array to remove user method to support hard_delete
 
 ## [10.3.1] - 2020-12-04
 
--- a/src/Api/Users.php
+++ b/src/Api/Users.php
@@ -181,13 +181,18 @@ class Users extends AbstractApi
     }
 
     /**
-     * @param int $id
+     * @param int   $id
+     * @param array $params {
+     *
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Thanks for this. Could you add a test and a change log entry please?

jeffersonmartin:
> Thanks for this. Could you add a test and a change log entry please?

I have added the Changelog entry. 

I've reviewed the existing tests and I think that the existing `shouldRemoveUser()` covers the existing test case, however I need some help with the right syntax to use.

https://github.com/GitLabPHP/Client/blob/10.3/tests/Api/UsersTest.php#L283-L298

I'm not sure on the best way to modify the test case to add the params array and how the `apiMock` function works. I've searched through several other tests trying to find a comparable test case but have only found `update` and `list` methods that don't do the same thing.

Here's the scaffolding of what I think we want to test:
1. Create a user
2. Create a group and add that user as an owner
3. Delete the user
4. Check if the group was also deleted

Any thoughts @GrahamCampbell? Thank you in advance!

GrahamCampbell:
Thanks for this. I'm just going to merge this, as is.

GrahamCampbell:
10.4.0 and 11.0.0 have now been released.

PR #615, opened 2021-01-12
[11.1] Add 'search_namespaces' for list-all-projects

cwandrey:
It must be possible to set this parameter for searching the namespace.
https://docs.gitlab.com/ee/api/projects.html#list-all-projects


GrahamCampbell:
Excellent, thanks.

PR #620, opened 2021-02-04
Added params to /releases call

MDXDave:
You can also use parameters like `per_page` and `page` on the `/releases` route.


GrahamCampbell:
It is not intended that someone should manually pass these arguments. Use the result pager abstraction. :)

PR #625, opened 2021-03-04
[11.3] Add support for a single inherited member

YKWeyer:
This adds support for reading the permissions of a specific member for a group/project including inherited permissions:
https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-members

I tried to respect the naming logic already used:

|                     | All members  | Single member |
|---------------------|--------------|---------------|
| Without inheritance | `members`    | `member`      |
| With inheritance    | `allMembers` | `allMember`   |

If you think this might be confusing, i could also name it `inheritedMember` (although this might not be perfect either, since both regular AND inherited members will be returned)

### DIFF
--- a/src/Api/Groups.php
+++ b/src/Api/Groups.php
@@ -158,6 +158,17 @@ class Groups extends AbstractApi
     /**
      * @param int|string $group_id
      * @param int        $user_id
+     *
+     * @return mixed
+     */
+    public function allMember($group_id, int $user_id)
+    {
+        return $this->get('groups/'.self::encodePath($group_id).'/members/all/'.self::encodePath($user_id));
+    }
+
+    /**
+     * @param int|string $group_id
+     * @param int        $user_id
      * @param int        $access_level
      *
      * @return mixed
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Thanks for this. Could you add some tests too?

YKWeyer:
:+1: I just added tests for the two `allMember` methods.

This might be caused by my setup, but `make install` wouldn't succeed entirely ("Couldn't find any bin namespace."), and I had to 
add `"phpunit/phpunit": "^9.0",` in the dev requirements of composer in order to run the tests.

GrahamCampbell:
Try running `make clean && make install && make test`. You may get some bin warnings, but they should be harmless.

GrahamCampbell:
If you have custom composer config in your home directory or globally installed packages, it's possible that could have messed things up.

GrahamCampbell:
Anyway, this looks good to me. Thanks. 🍻

PR #632, opened 2021-03-22
[11.4] Add parameters to the list of project repository tags

bbujisic:
Implementing the optional parameters from https://docs.gitlab.com/ee/api/tags.html#list-project-repository-tags

### DIFF
--- a/src/Api/Tags.php
+++ b/src/Api/Tags.php
@@ -18,12 +18,26 @@ class Tags extends AbstractApi
 {
     /**
      * @param int|string $project_id
+     * @param array      $parameters {
+     *
+     *     @var string $order_by Return tags ordered by `name` or `updated` fields. Default is `updated`.
+     *     @var string $sort     Return tags sorted in asc or desc order. Default is desc.
+     *     @var string $search   Return list of tags matching the search criteria. You can use `^term` and `term$` to
+     *                           find tags that begin and end with term respectively.
+     * }
      *
      * @return mixed
      */
-    public function all($project_id)
+    public function all($project_id, array $parameters = [])
     {
-        return $this->get($this->getProjectPath($project_id, 'repository/tags'));
+        $resolver = $this->createOptionsResolver();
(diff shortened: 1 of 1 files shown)


src/Api/Tags.php
```diff
@@ -82,4 +91,19 @@ public function updateRelease($project_id, string $tag_name, array $params = [])
     {
         return $this->put($this->getProjectPath($project_id, 'repository/tags/'.self::encodePath($tag_name).'/release'), $params);
     }
+
+    /**
+     * @return OptionsResolver
+     */
+    protected function createOptionsResolver(): OptionsResolver
```
glensc: it is not a very good idea to overwrite method with different purpose. another developer, already acquainted with this codebase, may not have idea that in this specific class $this-> createOptionsResolver() behaves differently than in all other classes.

indeed, they are able to discover it if they start troubleshooting, and find this difference, already by that time they are annoyed by the lost time and their motivation goes down.

suggestion: do not override the method, use local `$resolver` in the `all()` method. like rest of the similar code in this project. the `$resolver` does not have common options with any other method in this class.

glensc: if you insist on a separate method, perhaps `createOptionsResolverForMethodAll`, the first suggestion to use a name like `createOptionsResolverAll` will only create additional confusion as it's not clear is for "all" methods or just the fact that the method name is `all` :D

also, use `private`, not `protected` by default in new methods, if there's no known reason to use wider visibility.

bbujisic: I copied the solution from existing classes and existing `all()` methods. See `Issues::createOptionsResolver()` or `AbstractApi::createOptionsResolver()` for examples.

However, I agree with your argument that the override doesn't make sense as it won't be used anywhere else in the class and doesn't contribute the readability of the code.

The code is amended.

GrahamCampbell:
Thank you, both.

GrahamCampbell:
You can install this code using the version constraint `^11.4@dev`.

PR #635, opened 2021-03-30
[11.4] Add parameters to the list project issue discussions

danaildinev:
Implementing additional parameters for pagination: https://docs.gitlab.com/ce/api/#pagination
This way we can set parameter 'per_page' to custom value and get more results than the default value of 20.


GrahamCampbell:
You should use the result pager for this.

danaildinev-mms:
Yeah, it is working fine with ResultPager. Thank you!

PR #636, opened 2021-04-03
[11.5] Added support for filtering environments by state

tehplague:
Add support to limit environments by state (https://docs.gitlab.com/ee/api/environments.html#list-environments)


GrahamCampbell:
Thanks. :)

PR #642, opened 2021-04-14
[11.5] Added functions to change the active state for gitlab users

FSD-Christian-ADM:
- added function to activate user
- added function to deactivate user


GrahamCampbell (approved):
Thanks! 🍻 :shipit:

PR #644, opened 2021-05-16
[11.5] Add packages endpoints

rdrenth:
I've added the [Gitlab Packages](https://docs.gitlab.com/ee/api/packages.html) endpoints.

For Groups a `packages` method to list all packages within that group, the other methods are in the `Packages` class.


vekien:
Just commenting so this doesn't close, any ETA on if this will get merged?

GrahamCampbell:
Thanks. :)

PR #664, opened 2021-09-23
[11.5] Add support for locked merge requests

gallien-lbr:
According to GITLAB API documentation, there is a special state for "locked" merge requests.

https://docs.gitlab.com/ee/api/merge_requests.html


gallien-lbr:
I forgot in my original PR but It should be also added  in the merge request API optionResolver.

GrahamCampbell:
Please push that fix to the same branch, and it'll show up here too. :)

PR #667, opened 2021-10-21
[11.5] Allow using psr/cache 2.0

### LINKED ISSUE #647: Symfony 5.2+ PSR\Cache
It seem PSR 2.0 is now by default on new symfony app (5.2+) and bundle no more compatible with the PSR 2.0 version.

` m4tthumphrey/php-gitlab-api 11.5.x-dev requires psr/cache ^1.0 -> found psr/cache[1.0.0, 1.0.1] but the package is fixed to 2.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update `command.``

Any work in progress on this ?

mattjmattj:
Since version 1.7.2, php-http/cache-plugin can work with psr/cache 2.0 (this was the only difference in their changelog).
Here, we allow both versions, in order to remain compatible with PHP 7

Fixes #647


20uf (approved):

JanMikes:
Hi, could we add psr/cache 3.0 support as well? 

How can i help this to get merged?

GrahamCampbell:
> Hi, could we add psr/cache 3.0 support as well?

No. That is not possible.

GrahamCampbell:
Thanks. 🎄

quentint:
Hi, could you tell us a bit more why this would not be possible?

It is currently impossible to install `zeichen32/gitlabapibundle` (that relies on `m4tthumphrey/php-gitlab-api`) on a Symfony 6 app without forcing psr/cache on `^2.0`.

So `composer require zeichen32/gitlabapibundle` fails, but `composer require zeichen32/gitlabapibundle psr/cache:^2.0` works.

Thanks for your help!

quentint:
For those subscribed to this PR, see https://github.com/GitLabPHP/Client/releases/tag/11.6.0 😉

GrahamCampbell:
At that time, it was not possible. Now it is. All the upstream moving parts are in place. :)

PR #670, opened 2021-11-05
[11.8] Allow specifying params for Wiki::showAll()

spawnia:
https://docs.gitlab.com/ee/api/wikis.html#list-wiki-pages

### DIFF
--- a/src/Api/Wiki.php
+++ b/src/Api/Wiki.php
@@ -40,13 +40,19 @@ class Wiki extends AbstractApi
 
     /**
      * @param int|string          $project_id
-     * @param array<string,mixed> $params
+     * @param array<string,mixed> $params {
+     *     @var bool $with_content Include pages’ content
+     * }
      *
      * @return mixed
      */
     public function showAll($project_id, array $params)
     {
-        return $this->get($this->getProjectPath($project_id, 'wikis'), $params);
+        $resolver = $this->createOptionsResolver();
+        $resolver->setDefined('with_content')
+            ->setAllowedTypes('with_content', 'bool');
+
+        return $this->get($this->getProjectPath($project_id, 'wikis'), $resolver->resolve($params));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
You should use the result pager to fetch different pages, assuming that is your use case here.

spawnia:
Thanks for the hint, I will definitely keep that in mind as we probably have some wikis with over 20 pages 👍

spawnia:
Passing the option is still useful though:

```php
        $pages = (new ResultPager($this->gitlab->connection()))->fetchAll(
            $this->gitlab->wiki(),
            'showAll',
            [
                $projectId,
                ['with_contents' => true]
            ],
        );
```

spawnia:
Is there anything else I can do here? Would love to see this merged, I am currently blocked by being unable to use the GitLab API to its full extent.

20uf (approved):

src/Api/Wiki.php
```diff
@@ -17,8 +17,8 @@
 class Wiki extends AbstractApi
 {
     /**
-     * @param int|string $project_id
-     * @param array      $params
+     * @param int|string          $project_id
+     * @param array<string,mixed> $params
```
GrahamCampbell, line 21: Please adjust this doc to be similar to the other parameters doc, and use the options resolver.

GrahamCampbell (changes requested):

src/Api/Wiki.php
spawnia: Done.

spawnia:
I don't know what to do about the [test failure](https://github.com/GitLabPHP/Client/runs/4642485219?check_suite_focus=true), tests against the local PHP 7.4 setup (using `make`) work.

GrahamCampbell:
I was unable to push to your branch because you unchecked the option. I'm going to re-create this PR.

GrahamCampbell:
Thanks for the PR. :)

PR #675, opened 2021-11-23
[11.6] Added addShare method

peterjaap:


src/Api/Groups.php
```diff
@@ -205,6 +205,35 @@ public function removeMember($group_id, int $user_id)
     {
         return $this->delete('groups/'.self::encodePath($group_id).'/members/'.self::encodePath($user_id));
     }
+    
+    /**
+     * @param int|string $group_id
+     * @param array      $parameters
```
GrahamCampbell: Please add phpdoc for this. I cannot accept this PR without this. Please see the other methods for an example.

GrahamCampbell (changes requested):
Thanks for the PR. I've left some feedback inline.

src/Api/Groups.php
peterjaap: Done

GrahamCampbell:
Thanks. :)

PR #676, opened 2021-11-23
[11.6] Add parameters to `addVariable` on group level

peterjaap:


src/Api/Groups.php
```diff
@@ -383,10 +383,15 @@ public function variable($group_id, string $key)
      * @param string     $key
      * @param string     $value
      * @param bool|null  $protected
+     * @param array<string,mixed> $parameters        {
```
GrahamCampbell: Please follow the code standard from the rest of the codebase.

src/Api/Groups.php
```diff
@@ -383,10 +383,15 @@ public function variable($group_id, string $key)
...
+     *
+     *      @var string masked          true or false
```
GrahamCampbell: typo `masked` vs `$masked`

src/Api/Groups.php
```diff
@@ -396,6 +401,8 @@ public function addVariable($group_id, string $key, string $value, ?bool $protec
         if ($protected) {
             $payload['protected'] = $protected;
         }
+        
+        $payload = \array_merge($parameters, $payload);
```
GrahamCampbell: Please explicitly add the two supported parameters only.

GrahamCampbell (changes requested):
Thanks for the PR. I have left some inline comments.

GrahamCampbell:
Thanks. :)

PR #679, opened 2021-12-02
[11.6] Add support for protected tags

internalsystemerror:
Adds some missing endpoints regarding protected tags.

Closes https://github.com/GitLabPHP/Client/issues/678.

PR #680, opened 2021-12-13
[11.5] Add `deleteProtectedBranch` method to projects api

matt-h:
Handles the functionality to delete/unprotect a protected branch which is documented here:
https://docs.gitlab.com/ee/api/protected_branches.html#unprotect-repository-branches

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1190,6 +1190,17 @@ class Projects extends AbstractApi
 
     /**
      * @param int|string $project_id
+     * @param string     $branch
+     *
+     * @return mixed
+     */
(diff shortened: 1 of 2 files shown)


20uf (approved):

matt-h:
I put this together as I needed the function for a project I was working on, after submitting it I see that basically the same thing was also included in the PR: https://github.com/GitLabPHP/Client/pull/679

I didn't notice that one at first since it talks about Tags, but it included this branch change in the code as well. If that does get merged this can be skipped as redundant.

GrahamCampbell:
This PR is in good shape for merging, unlike the other one, so I'm going to merge this instead. The other one will need to be rebased. Thanks for sending this in.

PR #685, opened 2022-01-25
[11.8] Add parameters to `SystemHooks::create()`

edspc:
Hi. This implements params for the create method.
https://docs.gitlab.com/ee/api/system_hooks.html#add-new-system-hook


GrahamCampbell:
Thanks for the PR. I've made some adjustments, and merged. :)

PR #688, opened 2022-02-07
Improve handling of non-array json responses to prevent unexpected errors

vvasiloi:
I was implementing the [container registry](https://docs.gitlab.com/ee/api/container_registry.html) endpoints on top of this library when I ran into an unexpected error when decoding the response body.
The [delete registry repository](https://docs.gitlab.com/ee/api/container_registry.html#delete-registry-repository) endpoint returns a response with status `202 Accepted` and body `202`.
The current implementation doesn't accept integers as safe non-array JSON responses.
The proposed new implementation will return the response body as is if it's not deserialized into an array.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,9 @@
         }
     },
     "config": {
-        "preferred-install": "dist"
+        "preferred-install": "dist",
+        "allow-plugins": {
+            "bamarni/composer-bin-plugin": true
+        }
     }
 }
--- a/src/HttpClient/Message/ResponseMediator.php
+++ b/src/HttpClient/Message/ResponseMediator.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 
 namespace Gitlab\HttpClient\Message;
 
+use Gitlab\Exception\NonArrayJsonResponseException;
 use Gitlab\Exception\RuntimeException;
 use Gitlab\HttpClient\Util\JsonArray;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -57,6 +57,9 @@
         }
     },
     "config": {
-        "preferred-install": "dist"
+        "preferred-install": "dist",
+        "allow-plugins": {
+            "bamarni/composer-bin-plugin": true
```
vvasiloi, line 62: Not explicitly allowing plugins is deprecated and will be removed in July 2022.
https://getcomposer.org/doc/06-config.md#allow-plugins

src/HttpClient/Message/ResponseMediator.php
```diff
@@ -62,8 +63,16 @@ public static function getContent(ResponseInterface $response)
     {
         $body = (string) $response->getBody();
 
-        if (!\in_array($body, ['', 'null', 'true', 'false'], true) && 0 === \strpos($response->getHeaderLine(self::CONTENT_TYPE_HEADER), self::JSON_CONTENT_TYPE)) {
-            return JsonArray::decode($body);
+        if ('' === $body) {
+            return $body;
```
vvasiloi, line 67: Early return for cleaner execution path and district code coverage.

src/HttpClient/Util/JsonArray.php
```diff
@@ -39,8 +40,8 @@ public static function decode(string $json): array
             throw new RuntimeException(\sprintf('json_decode error: %s', \json_last_error_msg()));
         }
 
-        if (null === $data || !\is_array($data)) {
```
vvasiloi, line 42: The null check is redundant.

GrahamCampbell:
Thanks for the PR. I don't feel like this is the right solution. The mistake is that an attempt is made to decode the body. That line should be removed from the API class in question.

GrahamCampbell:
You'll want to mimic `getAsResponse` in your PR, for delete. :)

vvasiloi:
Thanks for the hint!

PR #689, opened 2022-02-10
drop patchlevel for php in composer.json

cwandrey:
You should do without the patch level of PHP in the composer.json. 
Many operating systems (for example Ubuntu 20.04 (PHP 7.4.3)) rarely support the latest PHP version. 
This repository does not necessarily require changes from the respective patch levels.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         }
     ],
     "require": {
-        "php": "^7.4.15 || ^8.0.2",
+        "php": "^7.4 || ^8.0",
         "ext-json": "*",
         "ext-xml": "*",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I am not interested in supporting very outdated versions, and PHP 8.0.0 and 8.0.1 don't work properly with this package (7.4.15 was released the same day as 8.0.2). If you are on a distro that has backported stuff, you can tell composer to emulate you being on 7.4.15. :)

GrahamCampbell:
In your app composer.json:

```json
    "config": {
        "platform": {
            "php": "7.4.15"
        }
    }
```

Issue #638, opened 2021-04-09
Pipeline creation fails due to invalid parameters


IAL32:
According to the documentation [here](https://docs.gitlab.com/ee/api/pipelines.html#create-a-new-pipeline), to create a pipeline this is the correct POST query:

`curl --request POST --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/projects/1/pipeline?ref=master"`

Currently though, the abstract API for the POST request does not handle correctly the `ref` query parameter:

```php
$response = $this->client->getHttpClient()->post(self::prepareUri($uri), $headers, $body);
```

As you can see, the `self::prepareUri($uri)` does not take into account the `ref` parameter from the `projects()->createPipeline` method.

Currently, `createPipeline` is the only place that I see that has this caveat.

IAL32:
Not an issue, I was just putting the parameters wrong.

IAL32:
Reopened, as the issue is still relevant, and a comprehensive PR is being made to fix it.

GrahamCampbell:
Hmmm, GitLab should really fix this on their side. This is inconsistent with how all their other API routes behave.

GrahamCampbell:
I'll merge your fix to unblock things, but I don't like it, lol. :)

PR #695, opened 2022-02-16
[11.8] Fixing `createPipeline` behavior to match API

IAL32:
This PR stems from https://github.com/GitLabPHP/Client/issues/638 and closes it.

Tests are also fixed to match actual behavior.

### DIFF
--- a/src/Api/AbstractApi.php
+++ b/src/Api/AbstractApi.php
@@ -106,7 +106,7 @@ abstract class AbstractApi
      *
      * @return mixed
      */
-    protected function post(string $uri, array $params = [], array $headers = [], array $files = [])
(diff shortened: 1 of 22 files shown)


src/Api/AbstractApi.php
```diff
@@ -106,7 +106,7 @@ protected function get(string $uri, array $params = [], array $headers = [])
      *
      * @return mixed
      */
-    protected function post(string $uri, array $params = [], array $headers = [], array $files = [])
+    protected function post(string $uri, array $params = [], array $headers = [], array $files = [], array $uriParams = [])
```
GrahamCampbell, line 109: Adding a parameter here is a major breaking change.

GrahamCampbell (changes requested):

PR #696, opened 2022-02-16
[11.9] Add support for Advanced Search API

IAL32:
The [Advanced Search API](https://docs.gitlab.com/ee/api/search.html#advanced-search-api) can be called as follows:

`(GET) /search`

Attribute | Type | Required | Description
-- | -- | -- | --
scope | string | yes | The scope to search in
search | string | yes | The search query
state | string | no | Filter by state. Issues and merge requests are supported; it is ignored for other scopes.
confidential | boolean | no | Filter by confidentiality. Issues scope is supported; it is ignored for other scopes.
order_by | string | no | Allowed values are created_at only. If this is not set, the results are either sorted by created_at in descending order for basic search, or by the most relevant documents when using advanced search.
sort | string | no | Allowed values are asc or desc only. If this is not set, the results are either sorted by created_at in descending order for basic search, or by the most relevant documents when using advanced search.

This PR adds support for this API.


src/Api/Projects.php
```diff
@@ -78,6 +78,8 @@ public function all(array $parameters = [])
             'id', 'name', 'path', 'created_at', 'updated_at', 'last_activity_at',
             'repository_size', 'storage_size', 'packages_size', 'wiki_size',
         ];
+        $resolver->setDefined('pagination');
```
GrahamCampbell, line 81: These parameters are not intentionally not supported. Please remove them. You need to use the result pager class.

GrahamCampbell (changes requested):

PR #698, opened 2022-02-18
[11.8] Allow chmod action and execute_filemode attribute

lothid:
Enable usage of chmod action and execute_filemode attribute in createCommit function: https://docs.gitlab.com/14.7/ee/api/commits.html#create-a-commit-with-multiple-files-and-actions


GrahamCampbell:
Thanks. 🍻

PR #699, opened 2022-02-21
[11.8] Add support for protected tags

lothid:
Support for interacting with protected tags: https://docs.gitlab.com/ee/api/protected_tags.html


GrahamCampbell:
Great, thanks.

PR #705, opened 2022-04-09
[11.9] Add custom attribute support to `Users`

igittigitt:
Added options to user lookup functions:
- "with_custom_attributes" (true|false) to also get users custom attributes in resulting array
- "custom_attributes" can contain an array of custom attributes to lookup users

Added functions:
- "customAttributes(int $id)" to get all custom attributes defined for the user
- "customAttribute(int $id, string $key)" to get single attribute by name
- "createCustomAttribute(int $id, string $key, string $value)" to add another attribute (key/value)
- "updateCustomAttribute(int $id, string $key, string $value)" to update (or create) an existing attribute
- "removeCustomAttribute(int $id, string $key)" to delete an custom attribute by key

### DIFF
--- a/src/Api/Users.php
+++ b/src/Api/Users.php
@@ -64,6 +64,13 @@ class Users extends AbstractApi
             ->setAllowedTypes('blocked', 'bool')
             ->setAllowedValues('blocked', true)
         ;
+        $resolver->setDefined('with_custom_attributes')
+            ->setAllowedTypes('with_custom_attributes', 'bool')
+            ->setAllowedValues('with_custom_attributes', true)
+        ;
+        $resolver->setDefined('custom_attributes')
+            ->setAllowedTypes('custom_attributes', 'array')
+        ;
 
         return $this->get('users', $resolver->resolve($parameters));
     }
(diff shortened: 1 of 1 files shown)


src/Api/Users.php
```diff
@@ -510,4 +517,66 @@ public function events(int $user_id, array $parameters = [])
...
+    /**
+     * @param int    $id
+     * @param string $key
+     * @param string $value
+     *
+     * @return mixed
+     */
+    public function customAttribute(int $id, string $key)
+    {
+        return $this->get('users/'.self::encodePath($id).'/custom_attributes/'.$key);
```
GrahamCampbell, line 540: you forgot to encode the key

src/Api/Users.php
```diff
@@ -510,4 +517,66 @@ public function events(int $user_id, array $parameters = [])
...
+     */
+    public function customAttributes(int $id)
+    {
+        return $this->get('users/'.self::encodePath($id).'/custom_attributes');
+    }
+
+    /**
+     * @param int    $id
+     * @param string $key
+     * @param string $value
```
GrahamCampbell, line 534: invalid phpdoc

GrahamCampbell (changes requested):
Thanks for the PR. I've left some feedback to be addressed.

PR #721, opened 2022-09-28
[11.10] Add support for `Packages::addGenericFile()`

rodrigoaguilera:
This an attempt to support the API call to add a generic file
https://docs.gitlab.com/ee/user/packages/generic_packages/index.html

The regular `->put()` call wasn't working because it adds multipart headers around the file and those are not stripped by gitlab. Something like:
```
--63347ec2cd1206.13859581
Content-Type: text/plain
Content-Disposition: form-data; name="executable"; filename="executable"
Content-Length: 11

echo hello

--63347ec2cd1206.13859581--
```
So I had to add the `putFile()` method.

The gitlab api handles multipart forms when you upload an avatar to a project (also a PUT call) but not in this case.

I tested it with a dummy project and so far it works.

### DIFF
--- a/src/Api/AbstractApi.php
+++ b/src/Api/AbstractApi.php
@@ -155,6 +155,28 @@ abstract class AbstractApi
 
     /**
      * @param string               $uri
+     * @param string               $file
+     * @param array<string,string> $headers
+     * @param array<string,mixed>  $uriParams
+     *
+     * @return mixed
+     */
+    protected function putFile(string $uri, string $file, array $headers = [], array $uriParams = [])
+    {
+        $resource = self::tryFopen($file, 'r');
+        $body = $this->client->getStreamFactory()->createStreamFromResource($resource);
+
+        if ($body->isReadable()) {
+            $headers = \array_merge([ResponseMediator::CONTENT_TYPE_HEADER => self::guessFileContentType($file)], $headers);
+        }
+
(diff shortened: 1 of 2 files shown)


rodrigoaguilera:
I would like for this to be reviewed by a maintainer.

GrahamCampbell (changes requested):
Thanks for the PR. Can you resolve the merge conflicts and revert the changes to the CHANGELOG, please?

CHANGELOG.md
```diff
@@ -5,25 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [11.9.0] - 2023-03-XX
```
GrahamCampbell: Please revert your changes to this file.

GrahamCampbell (changes requested):

CHANGELOG.md
rodrigoaguilera: I got the git remotes messed. Now is untouched

GrahamCampbell:
Great, thanks. This has missed the v11.9.0 tag, but will be included in v11.10.0 when that goes out, and can still be installed with dev stability using a version constraint like `^11.10@dev`, in the meanwhile.

PR #723, opened 2022-10-03
add 'updated_at' to 'order_by' for pipelines()

marcsimeon:
The Gitlab-Api (https://docs.gitlab.com/15.4/ee/api/pipelines.html#list-project-pipelines) states that pipelines can be ordered by 'updated_at' as well. It already works by allowing it here.


GrahamCampbell:
Thanks for the PR. It looks like this has already been applied as part of another PR, on the 11.9 branch. 🚀

GrahamCampbell:

PR #726, opened 2022-10-17
[12.0] Add support for prefixed URIs on base path

abelharisov:
Support custom gitlab installations with url like host/some/path/to/gitlab

### DIFF
--- a/src/Api/AbstractApi.php
+++ b/src/Api/AbstractApi.php
@@ -36,7 +36,7 @@ abstract class AbstractApi
      *
      * @var string
      */
-    private const URI_PREFIX = 'api/v4/';
+    private const URI_PREFIX = '/api/v4/';
 
     /**
      * The client instance.
--- a/src/Api/AbstractApi.php
+++ b/src/Api/AbstractApi.php
@@ -36,7 +36,7 @@ abstract class AbstractApi
      *
      * @var string
      */
-    private const URI_PREFIX = '/api/v4/';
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR. I feel that this is a breaking change, so I think we need to delay this until a v12.0.0 release.

src/Api/Projects.php
```diff
@@ -282,6 +282,17 @@ public function createTrigger($project_id, string $description)
...
         ]);
     }
+    
+     /**
+     * @param int|string $project_id
+     * @param int        $trigger_id
+     *
+     * @return mixed
+     */
+    public function removeTrigger($project_id, int $trigger_id)
```
GrahamCampbell: Please remove this change.

GrahamCampbell (changes requested):

src/Api/Projects.php
sklmx: fixed

GrahamCampbell:
not stale

GrahamCampbell:
Not stale

GrahamCampbell:
This PR is still in consideration for the next major version.

datawench:
Any thoughts on when this might become current? I happen to have a project that has a nonstandard self-hosted path (it's like `gitlab.myproject.com/gitlab`), and it's taken me a while to track down that this is why the `host` attribute of the request's URI gets hosed.

PR #729, opened 2022-10-28
Compatibility with php 8.1

omarlopesino:
Hi. After running a command that checks the php compatibility, looks like the gitlab client for PHp is already compatible with PHP 8.1, it is only needed to update the composer.json dependency.

Please may you check it and merge if everything is okay? thanks!


GrahamCampbell:
This PR doesn't do anything. Please read the docs on how `^` works. ;)

PR #731, opened 2022-11-01
[11.9] Add support for `access_level` in `Projects::createProjectAccessToken`

dododedodonl:
Add the `access_level` parameter, as specified in the [documentation](https://docs.gitlab.com/ee/api/project_access_tokens.html#create-a-project-access-token).

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1345,6 +1345,11 @@ class Projects extends AbstractApi
             })
         ;
 
+        $resolver->setDefined('access_level')
+            ->setAllowedTypes('access_level', 'int')
+            ->setAllowedValues('access_level', [10, 20, 30, 40])
+        ;
+
         $resolver->setDefined('expires_at')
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR. 👍

GrahamCampbell:
It looks like the tests are failing here. I've had to revert this.

dododedodonl:
@GrahamCampbell I'm not sure what the problem is, I do not use psalm myself and wrote the code quite some time ago. It looks like it has to do something with the type that is specified in the doctag. It looks correct to me though, as it uses an int in the testcase.

GrahamCampbell:
The failure was not psalm, but phpunit (at least, the psalm failure was not related to this PR).

GrahamCampbell:
https://github.com/GitLabPHP/Client/actions/runs/4343185802/jobs/7584915249

dododedodonl:
I am not that used to github, and missed the second action job. While fixing this I mixed up another local version of the sdk, and it was easier to create a new branch.

PR #734, opened 2023-01-10
[11.9] Add support for `Jobs::artifactByJobId`

donkidd:
I'd like to be able to download an artifact by the jobId, and since  I noticed that wasn't yet available in the code I would add it.. . 

https://docs.gitlab.com/ee/api/job_artifacts.html#download-a-single-artifact-file-by-job-id

Additional code and test added..


GrahamCampbell:
Thanks. 🎉

GrahamCampbell:

PR #735, opened 2023-01-25
[11.9] Add support for `Users::usersStarredProjects`

wimski:
Even-though there was already support for getting the user's project with a 'starred' parameter, the difference is subtle:

> Get a list of visible projects _**owned**_ by the given user. When accessed without authentication, only public projects are returned.

https://docs.gitlab.com/ee/api/projects.html#list-user-projects

> Get a list of visible projects _**starred**_ by the given user. When accessed without authentication, only public projects are returned.

https://docs.gitlab.com/ee/api/projects.html#list-projects-starred-by-a-user

### DIFF
--- a/src/Api/Users.php
+++ b/src/Api/Users.php
@@ -175,6 +175,88 @@ class Users extends AbstractApi
     }
 
     /**
+     * @param int   $id
+     * @param array $parameters {
+     *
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Great, thanks. 🎉

GrahamCampbell:

PR #737, opened 2023-02-05
[11.10] Added tags order_by support for version value

DevMoritz:
added version value for the order_by attribute in GitLab 15.4.
https://gitlab.com/gitlab-org/gitlab/-/merge_requests/95150


GrahamCampbell (changes requested):
Thanks for the PR. Please update the phpdoc and tests, also.

DevMoritz:
@GrahamCampbell the phpdoc was updated. What exactly should I add as a test?

PR #745, opened 2023-04-04
[11.9] Remove Trailing slash on project path

### LINKED ISSUE #739: Cannot upload project avatar: 'avatar is invalid'
[Our project calls `uploadAvatar()`](https://github.com/acquia/cli/blob/fae9193756f42a2b873b573327a6747067e96b66/src/Command/CodeStudio/CodeStudioCommandTrait.php#L253) after creating a GitLab project. I'm pretty sure this has worked in the past, but recently stopped working, throwing an `avatar is invalid` exception: https://github.com/acquia/cli/issues/1377

I did some debugging and it looks like a problem with this library. A curl request with the same avatar to the same API endpoint works fine:

`curl --location --request PUT 'https://code.acquia.com/api/v4/projects/1092' --header 'Authorization: Bearer [snip]' --form 'avatar=@drupal_icon.png'`

I suspect GitLab changed something in the expected file or PUT format that's breaking this call.

ajuchacko:
Fix: #739 

✅ Added Testcase
✅ All tests are passing on local


danepowell:
Tested and confirmed this fixes https://github.com/GitLabPHP/Client/issues/739 (and by extension, https://github.com/acquia/cli/issues/1377). Thanks!

GrahamCampbell:
Thanks for starting work on this. I feel the root cause is that the project method is not meant to be called with an empty string. I've fixed this, accordingly.

ajuchacko:
> Thanks for starting work on this. I feel the root cause is that the project method is not meant to be called with an empty string. I've fixed this, accordingly.

You're right, It makes sense 👍

PR #746, opened 2023-04-08
[11.11] Add `expires_at` parameter to `Groups::addMember()`

### LINKED ISSUE #738: Allow expire date in Groups::addMember()
The addMember() API has several extra parameters, like expiry date, that aren't supported.
Would be great to support it, thank you!

https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project

ajuchacko:
Fix: #738 

- [x] Updated test `ShouldaddMember` test
- [x] Moved/Made `ACCESS_LEVELS` const for project/group access levels
- [x] Updated phpdoc
- [x] All tests passing local

Hey, @GrahamCampbell I've added an optional parameter(of array type) to `addMember` method, I'd like to make another pr removing both `$user_id, $acess_level` from the method parameters, instead they will be passed inside the array like we're doing in other places to branch 12.0 once this gets merged

Thanks 😀

### DIFF
--- a/src/Api/Groups.php
+++ b/src/Api/Groups.php
@@ -208,15 +208,36 @@ class Groups extends AbstractApi
      * @param int|string $group_id
      * @param int        $user_id
      * @param int        $access_level
+     * @param array      $parameters
      *
      * @return mixed
      */
-    public function addMember($group_id, int $user_id, int $access_level)
+    public function addMember($group_id, int $user_id, int $access_level, array $parameters = [])
     {
-        return $this->post('groups/'.self::encodePath($group_id).'/members', [
+        $resolver = $this->createOptionsResolver();
+
+        $dateNormalizer = function (OptionsResolver $optionsResolver, \DateTimeInterface $date): string {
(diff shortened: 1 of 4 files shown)


src/Api/Groups.php
```diff
@@ -208,15 +208,32 @@ public function allMember($group_id, int $user_id)
...
     {
-        return $this->post('groups/'.self::encodePath($group_id).'/members', [
+        $dateNormalizer = function (OptionsResolver $optionsResolver, \DateTimeInterface $date): string {
+            return $date->format('Y-m-d');
+        };
+
+        $resolver = $this->createOptionsResolver()
+            ->setRequired(['user_id', 'access_level'])
+            ->setDefined('expires_at')
+            ->setAllowedTypes('user_id', 'int')
```
GrahamCampbell: User id and access level should not be permitted in the parameters array.

GrahamCampbell (changes requested):

GrahamCampbell:
Thanks.

PR #747, opened 2023-04-16
Updatge README.md with additional examples

lostcarpark:
The only example currently is to create a new project. That's good but I think a lot of people using the library only want to retrieve information about projects, and I found it tricky to figure out how to do that. I've added an example `show()` call to retrieve information about a project, which I found a little tricky, so hopefully including in the README will help someone.


GrahamCampbell:
The purpose of the example was to show one API call. I'd highly recommend jumping into the source code to see what other methods are available.

PR #748, opened 2023-04-20
[11.10] Add support for `Schedules::takeOwnership` and `Schedules::play`

FelixJacobi:
During using this library, I noticed that these two API calls are not implemented yet. Taking ownership is especially necessary to be able to edit a schedule originally created by another GitLab account.


GrahamCampbell:
Great, thanks. 🎉

PR #716, opened 2022-08-05
[11.10] Added include_retried option on pipeline jobs API call

glady:
Added include_retried option on pipeline jobs API call.
Introduced with GitLab 13.9 @see https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-jobs

PR #751, opened 2023-05-03
[11.11] Added `include_retried` option on to `Jobs::pipelineBridges`

glady:
Added include_retried option on pipeline jobs API call.
Introduced with GitLab 13.9 https://github.com/see https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-jobs

Same as #716 with added parameter to phpdoc


GrahamCampbell:
Thanks.

PR #752, opened 2023-05-05
[11.11] Add support for additional parameters in `Projects::forks`

fjgarlin:
The `forks` endpoint allows a handy `search` parameter, so define it as an option. 

See the options for forks endpoint: https://docs.gitlab.com/ee/api/projects.html#list-forks-of-a-project

With the following code:
```
      $forks = $gitlab_client->projects()->forks($project, [
        'search' => 'search_term',
      ]);
```

**Currently**: throws exception as parameter is not defined.
**After applying changes**: returns list of forks that match "search_term"

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1052,8 +1052,8 @@ class Projects extends AbstractApi
      *     @var bool               $with_issues_enabled         Limit by enabled issues feature
      *     @var bool               $with_merge_requests_enabled Limit by enabled merge requests feature
      *     @var int                $min_access_level            Limit by current user minimal access level
-     *     @var \DateTimeInterface $updated_before              Limit results to projects last updated before the specified time.
-     *     @var \DateTimeInterface $updated_after               Limit results to projects last updated after the specified time.
+     *     @var \DateTimeInterface $updated_before              limit results to projects last updated before the specified time
+     *     @var \DateTimeInterface $updated_after               limit results to projects last updated after the specified time
      *     @var bool               $with_custom_attributes      Include custom attributes in response
      * }
      *
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1035,13 +1035,17 @@ class Projects extends AbstractApi
 
     /**
      * @param int|string $project_id
-     * @param array      $parameters
+     * @param array      $parameters {
+     *
+     *     @var string $search         Return list of forks matching the search criteria (optional)
+     * }
      *
      * @return mixed
      */
     public function forks($project_id, array $parameters = [])
     {
         $resolver = $this->createOptionsResolver();
+        $resolver->setDefined('search');
 
         return $this->get($this->getProjectPath($project_id, 'forks'), $resolver->resolve($parameters));
     }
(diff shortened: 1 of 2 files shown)


src/Api/Projects.php
```diff
@@ -1035,13 +1035,17 @@ public function languages($project_id)
 
     /**
      * @param int|string $project_id
-     * @param array      $parameters
+     * @param array      $parameters {
+     *
+     *     @var string $search         Return list of forks matching the search criteria (optional)
```
GrahamCampbell: If we are going to document this, we need to document all the available parameters.

src/Api/Projects.php
```diff
@@ -1035,13 +1035,17 @@ public function languages($project_id)
...
+     * }
      *
      * @return mixed
      */
     public function forks($project_id, array $parameters = [])
     {
         $resolver = $this->createOptionsResolver();
+        $resolver->setDefined('search');
```
GrahamCampbell, line 1088: Shouldn't this be defined in the options resolver function?

GrahamCampbell (changes requested):
Thanks for the PR. Please address the inline comments and add test coverage.

src/Api/Projects.php
fjgarlin: I'm not too familiar with the code inside the library, but the rest of the methods seem to be defining the options in the same way, within the method.

src/Api/Projects.php
fjgarlin: I can try that. I wanted to get the minimum PR that would work and then have it reviewed by maintainers (aka you) to be sure that this was the right approach. I'll try to add and document the other options.

src/Api/Projects.php
fjgarlin: I've done it now based on the current code (I took the `all` method as reference) and the information found here: https://docs.gitlab.com/ee/api/projects.html#list-forks-of-a-project

src/Api/Projects.php
fjgarlin: I've done it now based on the current code (I took the `all` method as a reference) and the information I found here: https://docs.gitlab.com/ee/api/projects.html#list-forks-of-a-project

src/Api/Projects.php
fjgarlin: I added all the other options available.

fjgarlin:
Inline comments were addressed and test coverage was added. Pleae review.

src/Api/Projects.php
GrahamCampbell: Yeh, this is not correct. Please look over the `createOptionsResolver` method in this class, and only add the things not already in there, or if search is always applicable, instead, add it there.

GrahamCampbell (changes requested):

src/Api/Projects.php
fjgarlin: I don't think that `search` should bubble up to the `createOptionsResolver` as that's used by many other methods. Note that there is no override for `createOptionsResolver` in this class, it is using the one of `AbstractApi.php` file.

The line `$resolver->setDefined('search');` was taken from the `all` method within this same class.

Not sure what more changes are required? I might need some more details.

GrahamCampbell:
Thanks.

PR #753, opened 2023-05-05
[11.11] Add support for additional parameters in `Projects::deployment`

alexdmccabe:
When getting deployments from projects, adding any of the options listed on https://docs.gitlab.com/ee/api/deployments.html#list-project-deployments results in an error like:

```
In OptionsResolver.php line 871: The option "status" does not exist. Defined options are: "page", "per_page". 
```

This adds the missing parameters.


GrahamCampbell:
Great, thanks!

PR #755, opened 2023-05-17
[11.11] Add support for `MergeRequests::showParticipants`

fjgarlin:
The merge_requests endpoint allows to bring participants as seen here https://docs.gitlab.com/ee/api/merge_requests.html#get-single-merge-request-participants


GrahamCampbell:
Thanks!

PR #757, opened 2023-06-07
[11.12] Passing parameters to removeVariable method

janfejtek:
With this change, it is possible to pass filter to the method https://docs.gitlab.com/ee/api/project_level_variables.html#delete-a-variable. 

This should keep it universal in future, but maybe separate filter parameter would be also possible

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1203,9 +1203,9 @@ class Projects extends AbstractApi
     /**
      * @param int|string           $project_id
      * @param string               $key
-     * @param array<string, mixed> $parameters {
+     * @param array<string, mixed> $parameters    {
      *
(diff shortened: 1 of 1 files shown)


GrahamCampbell (changes requested):
Thanks for the PR. Please add documentation to what parameters can be passed, in the usual way, and enforce with the options resolver.

janfejtek:
@GrahamCampbell I changed it, so it is under filter key to more follow standard used in other methods. But after that, I could not find any way to use resolver to validate keys on seconds level. 

Also the style check fails, but it looks like its issue with formating used in whole file, not only my change and looking at it, it follows project phpdoc fomatting

GrahamCampbell:
Thanks.

PR #758, opened 2023-06-07
[11.11] Add support for `tier` in `Environments::create`

janfejtek:
https://docs.gitlab.com/ee/api/environments.html#create-a-new-environment


GrahamCampbell:
Thanks!

PR #759, opened 2023-06-12
Proposal add project integrations slack / jira

coqmos:


pkolmann:
@coqmos I just found this and wanted to start on my own. Thanks for your work.

If I look at the GitLab Integrations API (https://docs.gitlab.com/ee/api/integrations.html) I think that your approach will blow up the projects.php file.

What do you think about an approach where you have an integrations() function in projects to get the overview, but have an extra Integrations class where we could handle the different integrations?

coqmos:
@pkolmann you are right it would I like your suggestion and will try to update the merge request later on :)

pkolmann:
@coqmos since I already had most if it ready, I took the liberty and added jira to my changes and did another pull request.

Please take a look and help me improve the code.

GrahamCampbell:
Closing in favour of #760.

PR #762, opened 2023-06-19
[11.11] Add support for `Users::removeUserIdentity`

pkolmann:
Add missing API to remove an external Identity mapping from a user

https://docs.gitlab.com/ee/api/users.html#delete-authentication-identity-from-user

### DIFF
--- a/src/Api/Users.php
+++ b/src/Api/Users.php
@@ -603,6 +603,6 @@ class Users extends AbstractApi
      */
     public function removeUserIdentity(int $user_id, string $provider)
     {
-        return $this->delete('users/'.self::encodePath($user_id).'/identities/'.$provider);
+        return $this->delete('users/'.self::encodePath($user_id).'/identities/'.self::encodePath($provider));
     }
 }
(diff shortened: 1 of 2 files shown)


src/Api/Users.php
```diff
@@ -592,4 +592,16 @@ public function events(int $user_id, array $parameters = [])
...
+    /**
+     * Deletes a user’s authentication identity using the provider name associated with that identity.
+     *
+     * @param int $user_id
+     * @param string $provider
+     * @return mixed
+     */
+    public function removeUserIdentity(int $user_id, string $provider)
+    {
+        return $this->delete('users/'.self::encodePath($user_id).'/identities/'.$provider);
```
GrahamCampbell: provider needs to be passed through `encodePath`

GrahamCampbell (changes requested):

src/Api/Users.php
pkolmann: Thanks for the hint. fixed.

GrahamCampbell:
Thanks.

PR #765, opened 2023-08-03
[11.12] Add support for additional parameters in `Projects::labels` and `Groups::labels`

rfussien:
Hello,
Here is a small contribution to support labels parameters. No BC, so I figured the 11.11 would do.
Not sure changing the test would make sens.


GrahamCampbell:
Thanks. This is a new feature, so needs to target 11.12, as per semver.

GrahamCampbell:
Thanks.

PR #767, opened 2023-08-10
[11.12] Add support for `filter` in `Projects::variable`

fului:
This PR implements the option to add a filter[environment_scope] 
to a variable request, when multiple variables
with the same name exist, but have different
environment scopes.

See documentation for more information
https://docs.gitlab.com/ee/api/project_level_variables.html#get-a-single-variable

Example on how to use
```php
$client->projects()->variable(
    $project_id,
    'host',
    [
        'filter' => [
            'environment_scope' => 'production'
        ]
    ]
);

return $variable;
```

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1214,15 +1214,13 @@ class Projects extends AbstractApi
     /**
      * @param int|string $project_id
      * @param string     $key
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks.

PR #769, opened 2023-08-11
[11.13] Addition of Status & Environment to list of possible values

### LINKED ISSUE #768: Add support for additional filters in the Deployment API
In https://docs.gitlab.com/ee/api/deployments.html there are additional filters available and I'd like to use at least a few of them...  

I'll submit a MR shortly with the updates I think are needed...

donkidd:
Added Status an Environment as parameters that can be sent to deployment method.  This is to fix #768

### DIFF
--- a/src/Api/Deployments.php
+++ b/src/Api/Deployments.php
@@ -24,7 +24,7 @@ class Deployments extends AbstractApi
      *                                              or ref fields (default is id)
      *     @var string $sort                        Return deployments sorted in asc or desc order (default is desc)
      *     @var string $status                      Return deployments filtered by status of deployment allowed
-+    *                                              values of status are 'created', 'running', 'success', 'failed',
+     *                                              values of status are 'created', 'running', 'success', 'failed',
      *                                              'canceled', 'blocked'
      *     @var string $environment                 Return deployments filtered to a particular environment
      * }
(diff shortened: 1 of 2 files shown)


src/Api/Deployments.php
```diff
@@ -36,6 +36,11 @@ public function all($project_id, array $parameters = [])
         $resolver->setDefined('sort')
             ->setAllowedTypes('sort', 'string')
             ->setAllowedValues('sort', ['desc', 'asc']);
+        $resolver->setDefined('status')
```
GrahamCampbell, line 43: Please also update the phpdoc

GrahamCampbell (changes requested):
Thanks for the PR. I've left notes inline to be addressed before merge.

src/Api/Deployments.php
donkidd: Ohh yea forgot about that, have updated the phpdoc with new values

donkidd: Finally got Style to pass, I kept missing a space in my phpdoc...

GrahamCampbell:
r.e. StyleCI, it gives you the diff with the fixes to download and then apply using git, so you don't have to take multiple tries. ;)

---

image

https://github.styleci.io/analyses/YOdYGj/diff

GrahamCampbell:
Thanks.

PR #772, opened 2023-08-21
[11.14] GraphQL API

### LINKED ISSUE #766: GraphQL API
Gitlab supports GraphQL: https://docs.gitlab.com/ee/api/graphql/

Github equivalent of this feature in the respective PHP SDK: https://github.com/KnpLabs/php-github-api/blob/master/lib/Github/Api/GraphQL.php

claudiu-cristea:
Fixes #766

### DIFF
--- a/src/Api/AbstractApi.php
+++ b/src/Api/AbstractApi.php
@@ -36,7 +36,7 @@ abstract class AbstractApi
      *
      * @var string
      */
-    private const URI_PREFIX = '/api/v4/';
+    protected const URI_PREFIX = '/api/v4/';
 
     /**
      * The access levels for groups and projects
--- a/src/Api/AbstractApi.php
+++ b/src/Api/AbstractApi.php
@@ -264,7 +264,7 @@ abstract class AbstractApi
             return null !== $value;
         });
 
-        return \sprintf('%s%s%s', self::URI_PREFIX, $uri, QueryStringBuilder::build($query));
+        return \sprintf('%s%s%s', static::URI_PREFIX, $uri, QueryStringBuilder::build($query));
     }
 
     /**
--- a/src/Api/GraphQL.php
+++ b/src/Api/GraphQL.php
@@ -19,7 +19,7 @@ class GraphQL extends AbstractApi
     /**
      * {@inheritDoc}
      */
-    protected const URI_PREFIX = '/api/';
+    protected const URI_PREFIX = '/api';
 
     /**
      * @param string $query
--- a/src/Api/GraphQL.php
+++ b/src/Api/GraphQL.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Gitlab API library.
+ *
+ * (c) Matt Humphrey <matth@windsor-telecom.co.uk>
+ * (c) Graham Campbell <hello@gjcampbell.co.uk>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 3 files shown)


src/Api/AbstractApi.php
```diff
@@ -36,7 +36,7 @@ abstract class AbstractApi
      *
      * @var string
      */
-    private const URI_PREFIX = '/api/v4/';
+    protected const URI_PREFIX = '/api/v4/';
```
dkarlovi, line 39: Do you really need to change this? Since you're using static now, it should automatically use the one from the child class, no? 🤔

claudiu-cristea, line 39: Maybe we should split version out of `URI_PREFIX`, into a protected class? Then we override the version in GraphQL as GitLab GraphQL is versionless

src/Api/GraphQL.php
```diff
@@ -0,0 +1,52 @@
...
+ * (c) Matt Humphrey <matth@windsor-telecom.co.uk>
+ * (c) Graham Campbell <hello@gjcampbell.co.uk>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gitlab\Api;
+
+class GraphQL extends AbstractApi
```
dkarlovi, line 17: Should have a few test cases.

dkarlovi (changes requested):

src/Api/AbstractApi.php
claudiu-cristea: In PHP `private` means a sub-class cannot extend/override it.

src/Api/GraphQL.php
claudiu-cristea: We should. But I have no time to work on

src/Api/AbstractApi.php
```diff
@@ -264,7 +264,7 @@ private static function prepareUri(string $uri, array $query = []): string
             return null !== $value;
         });
 
-        return \sprintf('%s%s%s', self::URI_PREFIX, $uri, QueryStringBuilder::build($query));
+        return \sprintf('%s%s%s', static::URI_PREFIX, $uri, QueryStringBuilder::build($query));
```
GrahamCampbell, line 267: Please revert these changes.

GrahamCampbell (changes requested):

src/Api/AbstractApi.php
claudiu-cristea: I don't think will work. GraphQL needs a different URL prefix

src/Api/AbstractApi.php
GrahamCampbell: why does it need to be overwritten? the changes you want don't need to re-use this?

claudiu-cristea: @GrahamCampbell, GraphQL is _versionless_. The URL prefix is `/api` instead of `/api/v4`, so `AbstractApi::prepareUriP()`, needs to get `/api` when subclassed as `\Gitlab\Api\GraphQL`

GrahamCampbell (changes requested):
I think it's a mistake to extend `AbstractApi` at all.

claudiu-cristea:
> I think it's a mistake to extend `AbstractApi` at all.

Maybe but I wanted to reuse `AbstractApi::post()`. Or should we move that in a trait?

GrahamCampbell:
We should just take the hit on a little duplication. Traits are horrible for this type of thing (most things imo :trollface:).

GrahamCampbell:
Closing due to inactivity.

PR #773, opened 2023-09-08
[11.11] Fix double encoding of job name in artifacts download

wilkolazki:
I found a bug where methods `Jobs::shouldGetArtifactsByRefName` and `Jobs::shouldGetArtifactByRefName` unnecesarily call `self::encodePath()` to encode query parameters, while all the query parameters are later encoded again in `AbstractApi::prepareUri` by call to `QueryStringBuilder::build()`. 

I've corrected those two methods and modified tests acordingly. I have not checked other methods for similar bugs, but I think it is probable it is there.


GrahamCampbell:
Thanks.

PR #777, opened 2023-10-03
[11.12] Add support for `author` in `Repositories::commits`

### LINKED ISSUE #776: [BUG] Missing author parameter on repositories()->commits()
As documentation states, there's an author parameter:
https://docs.gitlab.com/ee/api/commits.html

But this SDK is missing it:
https://github.com/GitLabPHP/Client/blob/11.11/src/Api/Repositories.php#L203

When i try to use it, it throws an exception saying that `author` doesn't exist.

```php
$client->repositories()->commits($projectId, [
    "author" => "demo@demo.com"
]);
```

DIrect curl requests against /repositories/commits with `author` parameter do work, so it's just missing in this sdk definition :o

tetreum:
Closes https://github.com/GitLabPHP/Client/issues/776

### DIFF
--- a/src/Api/Repositories.php
+++ b/src/Api/Repositories.php
@@ -212,6 +212,7 @@ class Repositories extends AbstractApi
 
         $resolver->setDefined('path');
         $resolver->setDefined('ref_name');
+        $resolver->setDefined('author');
         $resolver->setDefined('since')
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks.

PR #778, opened 2023-10-12
[11.14] Add project vulnerabilities endpoint

cwandrey:
https://docs.gitlab.com/ee/api/project_vulnerabilities.html#list-project-vulnerabilities

### DIFF
--- a/src/Api/Projects.php
+++ b/src/Api/Projects.php
@@ -1783,7 +1783,7 @@ class Projects extends AbstractApi
     }
 
     /**
-     * @param $project_id
+     * @param int|string $project_id
      * @param array $parameters
      *
      * @return mixed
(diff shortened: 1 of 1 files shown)


src/Api/Projects.php
```diff
@@ -1781,4 +1781,15 @@ public function deleteProtectedTag($project_id, string $tag_name)
     {
         return $this->delete($this->getProjectPath($project_id, 'protected_tags/'.self::encodePath($tag_name)));
     }
+
+    /**
+     * @param int|string $project_id
+     * @param array $parameters
```
GrahamCampbell, line 1787: We need test coverage and validation on this parameter.

GrahamCampbell (changes requested):
Thanks for the PR, but is not yet mergable without more changes, noted inline.

PR #780, opened 2023-10-25
[11.14] Add v4 Releases API

tomscholz:
Hi there,

I've made some improvements to the GitLab library:

1. Moved Releases API calls to their own dedicated file.
1. Separated Releases requests from Tags.

I've also copied over the relevant tests for the Tags implementation.

What needs to be done, to get this merged?

### DIFF
--- a/src/Api/Releases.php
+++ b/src/Api/Releases.php
@@ -1,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Gitlab API library.
+ *
+ * (c) Matt Humphrey <matth@windsor-telecom.co.uk>
+ * (c) Graham Campbell <hello@gjcampbell.co.uk>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gitlab\Api;
+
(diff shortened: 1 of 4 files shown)


src/Api/Releases.php
```diff
@@ -0,0 +1,96 @@
...
+
+class Releases extends AbstractApi
+{
+    /**
+     * @param int|string $project_id
+     * @param array $parameters {
+     *
+     * @return mixed
+     *
+     * @var string $sort Return releases sorted in asc or desc order. Default is desc.
```
GrahamCampbell, line 25: The syntax for this is wrong, and some of the other paramters docs are missing/incorrect, too.

tests/Api/TagsTest.php
```diff
@@ -99,6 +99,8 @@ public function shouldRemoveTag(): void
      *
      * @dataProvider releaseDataProvider
      *
+     * @deprecated
```
GrahamCampbell, line 102: don't mark tests as deprecated please

GrahamCampbell (changes requested):
Great start here. I've left some note inline that will need to be addressed before this PR can be merged.

GrahamCampbell:
Closing due to inactivity.

PR #782, opened 2023-10-31
[11.15] Repository branch and file check existence methods

misantron:


misantron:
Hello, take a look please

GrahamCampbell (changes requested):
These changes don't seem work properly. Instead of returning false if not exists, this actually throw an exception. We are, critically, missing tests for that case.

misantron:
```
try {
   $this->head($uri, []);
            
   return true;
} catch (Gitlab\Exception\RuntimeException $ex) {
   if (404 === $ex->getCode()) {
      return false;
   }
   throw $ex;
}
 ```
 @GrahamCampbell can I use such a code to handle exception and check entity existence?

misantron:
the question is still actual

PR #784, opened 2023-11-17
[11.14] Add users/personal_access_tokens methods

olivbd:


.gitignore
```diff
@@ -3,3 +3,4 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
+/composer.phar
```
GrahamCampbell, line 6: Please revert this change.

GrahamCampbell: It is intended that one uses the Makefile.

src/Api/Users.php
```diff
@@ -553,6 +553,62 @@ public function removeImpersonationToken(int $user_id, int $impersonation_token_
         return $this->delete('users/'.self::encodePath($user_id).'/impersonation_tokens/'.self::encodePath($impersonation_token_id));
     }
 
+    /**
+     * @param int   $user_id
+     * @param array $params
```
GrahamCampbell, line 558: We need documentation on the values that go in here - I know not all the code has this, but all new code needs it.

src/Api/Users.php
```diff
@@ -553,6 +553,62 @@ public function removeImpersonationToken(int $user_id, int $impersonation_token_
...
+
+    /**
+     * @param int         $user_id
+     * @param string      $name
+     * @param array       $scopes
+     * @param string|null $expires_at
+     *
+     * @return mixed
+     */
+    public function createPersonalAccessToken(int $user_id, string $name, array $scopes, ?string $expires_at = null)
```
GrahamCampbell, line 592: should expires at really be a string? how does a user know what to put here?

src/Api/Users.php
```diff
@@ -553,6 +553,62 @@ public function removeImpersonationToken(int $user_id, int $impersonation_token_
...
+     */
+    public function userPersonalAccessToken(int $user_id, int $personal_access_token_id)
+    {
+        return $this->get('users/'.self::encodePath($user_id).'/personal_access_tokens/'.self::encodePath($personal_access_token_id));
+    }
+
+    /**
+     * @param int         $user_id
+     * @param string      $name
+     * @param array       $scopes
```
GrahamCampbell, line 587: please document the array-value type

GrahamCampbell (changes requested):

src/Api/Users.php
GrahamCampbell: We almost certainly want to take a DateTimeImmutable here, and probably to use a params array, like in `Groups::issues`.

GrahamCampbell:
Closing due to inactivity.

PR #786, opened 2023-11-25
[11.14] Group milestones API: able to call GitLab API with addtional parameters updated_before and updated_after

wenchaoxue:
[List group milestones new parameters](https://docs.gitlab.com/ee/api/group_milestones.html#list-group-milestones)

```
updated_before:
Return only milestones updated before the given datetime. Expected in ISO 8601 format (2019-03-15T08:00:00Z).
Introduced in GitLab 15.10

updated_after:
Return only milestones updated after the given datetime. Expected in ISO 8601 format (2019-03-15T08:00:00Z).
Introduced in GitLab 15.10
```

 - call Group milestones API with addtional parameters updated_before and updated_after and add unit tests
 - add missing unit tests on the parameters iids, state, search

### DIFF
--- a/src/Api/GroupsMilestones.php
+++ b/src/Api/GroupsMilestones.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 
 namespace Gitlab\Api;
 
+use Symfony\Component\OptionsResolver\Options;
+
 class GroupsMilestones extends AbstractApi
 {
     /**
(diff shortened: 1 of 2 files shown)


src/Api/GroupsMilestones.php
```diff
@@ -32,7 +32,11 @@ class GroupsMilestones extends AbstractApi
      *
      *     @var int[]  $iids   return only the milestones having the given iids
      *     @var string $state  return only active or closed milestones
-     *     @var string $search Return only milestones with a title or description matching the provided string.
+     *     @var string $search Return only milestones with a title or description matching the provided string
+     *     @var string $updated_before Return only milestones updated before the given datetime
+     *                                 Expected in ISO 8601 format (2019-03-15T08:00:00Z). Introduced in GitLab 15.10
+     *     @var string $updated_after  Return only milestones updated after the given datetime
```
GrahamCampbell: Please take `DateTimeInterface` here rather than a string, just like in `Groups::issues`.

GrahamCampbell (changes requested):
Thanks for making a start here. There is just one change to be made.

src/Api/GroupsMilestones.php
wenchaoxue: Sorry for the delay, finally I have time to fix it - [here](https://github.com/GitLabPHP/Client/pull/786/commits/f35c0a180d7f545a55fded709a2d5e696ddf13a2)

Use `DateTimeInterface` is much better than string.

wenchaoxue: @GrahamCampbell the PR is ready to be reviewed in case you haven't received this notice before. Thanks for reviewing it.

GrahamCampbell:
Thanks.

PR #787, opened 2023-12-14
[11.14] Update `MergeRequests::all` to use millisecond precision

janedbal:
We are iteratively fetching new MRs via:
```
updated_after=2023-01-24T18:05:25.801Z
&sort=asc
&order_by=updated_at
```

And there is more than 100 MR updated within a second. Current implementation causes the same MRs to be analysed again every time. But the API seems to support microsecond precision, so this should fix it.

### DIFF
--- a/src/Api/MergeRequests.php
+++ b/src/Api/MergeRequests.php
@@ -74,7 +74,7 @@ class MergeRequests extends AbstractApi
     {
         $resolver = $this->createOptionsResolver();
         $datetimeNormalizer = function (Options $resolver, \DateTimeInterface $value): string {
-            return $value->format('c.u');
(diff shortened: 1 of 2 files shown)


janedbal:
Adjusted the formatting:

- `2023-01-24T18:05:25.801Z` works
- `2023-01-24T18:05:25.801+00:00` does not work

GrahamCampbell:
This uses millisecond precision, not microsecond?

janedbal:
Yes, adjusted title. GitLab API returns datetimes with milliseconds, not micro. So that is probably the best precision we can use.

janedbal:
@GrahamCampbell any needed changes here?

janedbal:
ping

GrahamCampbell:
Thanks.

PR #788, opened 2023-12-14
[11.4] Fix project events not allowing approved action

TZK-:
Add the `approved` action type in project events according to the [documentation](https://docs.gitlab.com/ee/user/profile/contributions_calendar.html#user-contribution-events)


GrahamCampbell:
Thanks.

PR #789, opened 2024-01-12
[11.15] Add support for listing merge requests associated with a commit

zluiten:
Api specification: https://docs.gitlab.com/ee/api/commits.html#list-merge-requests-associated-with-a-commit

### DIFF
--- a/src/Api/Repositories.php
+++ b/src/Api/Repositories.php
@@ -270,6 +270,19 @@ class Repositories extends AbstractApi
 
     /**
      * @param int|string $project_id
+     * @param string     $sha
+     *
+     * @return mixed
+     */
+    public function commitMergeRequests($project_id, string $sha)
+    {
+        return $this->get(
+            $this->getProjectPath($project_id, 'repository/commits/'.self::encodePath($sha).'/merge_requests'),
+        );
+    }
+
+    /**
+     * @param int|string $project_id
      * @param array      $parameters {
      *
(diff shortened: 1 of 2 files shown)


src/Api/Repositories.php
```diff
@@ -268,6 +268,23 @@ public function commitRefs($project_id, string $sha, array $parameters = [])
...
     }
 
+    /**
+     * @param int|string $project_id
+     * @param string     $sha
+     * @param array      $parameters
+     *
+     * @return mixed
+     */
+    public function commitMergeRequests($project_id, string $sha, array $parameters = [])
```
GrahamCampbell: We need param documentation and validation for the supported parameters please.

GrahamCampbell (changes requested):
Thanks for getting this started. I've left a note inline to be addressed. I know existing code is not always with this doc, but net new code needs it.

src/Api/Repositories.php
zluiten: I've removed the "parameters" parameter as this api endpoint does not seem to support any parameters.

zluiten: @GrahamCampbell

zluiten:
> This issue has been automatically marked as stale because there has been no recent activity. It will be closed after 30 days if no further activity occurs. Thank you for your contributions.

Don't close

zluiten:
@GrahamCampbell what's up with the closing?

PR #791, opened 2024-02-19
[11.15] feat: stars on projects

Cerdic:
methods star() to star a project and starrers() to get the list of project starrers


GrahamCampbell (changes requested):

PR #792, opened 2024-02-19
[11.15] feat: Context Commits on merge requests

Cerdic:
methods contextCommits() to list context, and addContextCommits() to append context commits


GrahamCampbell (changes requested):

PR #793, opened 2024-02-19
[11.15] feat: award emojis on issues

Cerdic:
methods addAwardEmoji() to append an emoji on an issue, and showNoteAwardEmoji() and addNoteAwardEmoji() to see and append emoji on comments on an issue


GrahamCampbell (changes requested):

PR #805, opened 2024-04-24
Implement stop stale environments method

wojcikmariusz:
Implement environment stop stale method: https://docs.gitlab.com/ee/api/environments.html#stop-stale-environments


GrahamCampbell:
Replaced by https://github.com/GitLabPHP/Client/pull/879.

PR #819, opened 2024-09-30
Fix implicit nullable types to avoid PHP 8.4 warnings

eiriksm:


danepowell:
Thanks for a good start, but this is incomplete. See https://github.com/GitLabPHP/Client/pull/822

GrahamCampbell:
Thanks for the PR. Version 12.0 will provide PHP 8.4 compatibility.

PR #822, opened 2024-11-26
PHP 8.4 compatibility

danepowell:
@GrahamCampbell would you mind reviewing this?

### DIFF
--- a/src/Api/Groups.php
+++ b/src/Api/Groups.php
@@ -91,7 +91,7 @@ class Groups extends AbstractApi
      *
      * @return mixed
      */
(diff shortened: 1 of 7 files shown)


danepowell:
@GrahamCampbell do you think you'll have a chance to review this or should we fork this library if we need PHP 8.4 support?

TomCan:
Any updates on this one? We're currently running into issues as well after a PHP upgrade and we'de rather not have to fork.

jissereitsma:
I tested out this PR using a composer patch and it works for me nicely. It would be great if it could be merged!

GrahamCampbell:
Thanks for the PR. Version 12.0 will provide PHP 8.4 compatibility.

PR #823, opened 2025-01-17
feat: support active/inactive state for listing project access tokens

beschoenen:
This PR adds support for filtering project access tokens. With the same API as deploy tokens.

ref: https://docs.gitlab.com/ee/api/project_access_tokens.html


beschoenen:
Rebasing to 12.0

Issue #832, opened 2025-03-21
Create Pipeline variables are not set properly


ivancli:
The GitLab API documentation specifies that pipeline variables should be included directly in the request body 
(see: https://docs.gitlab.com/api/pipelines/#create-a-new-pipeline). 

However, the `createPipeline()` method in the GitLab PHP Client incorrectly places these variables within a variables field in the request body, leading to a `variables is invalid` error from the GitLab API.
(see: https://github.com/GitLabPHP/Client/blob/362450ff932296bc15796f97da8cf57941650037/src/Api/Projects.php#L355)

PR #833, opened 2025-03-21
#832 - set variables to parameters in createPipeline

ivancli:
This is to resolve issue #832 by setting variables to $parameters directly


GrahamCampbell:
Replaced with https://github.com/GitLabPHP/Client/pull/891.

PR #838, opened 2025-07-04
Add way to update a deploy key on a project.

niconoe-:
This pull request propose you to add a way to update a deploy key assigned to a project.

Current methods can be used to list, add, enable, disable or delete a deploy key, but no method was designed to update a deploy key, which makes sense when the deploy key is already created and one wants to change its name or grant access to push.

Related documentation: https://docs.gitlab.com/api/deploy_keys/#update-deploy-key


niconoe-:
I'm uping it because I don't wan't to see this closed. If any modifications are expected, please tell me so :wink:

GrahamCampbell:
Replaced with https://github.com/GitLabPHP/Client/pull/892.

PR #842, opened 2025-08-20
feat: Added latest pipeline route

wbsartori:
This pull request proposes that you add the route to query the last pipeline that was executed for the project being requested.

https://docs.gitlab.com/api/pipelines/#get-the-latest-pipeline


GrahamCampbell:
Thanks for getting started here. I've corrected this, replacing it with https://github.com/GitLabPHP/Client/pull/897.

PR #843, opened 2025-08-29
Add support for listing merge requests associated with a commit

zluiten:
Api specification: https://docs.gitlab.com/ee/api/commits.html#list-merge-requests-associated-with-a-commit


zluiten:
I think there is still value in adding this

GrahamCampbell:
Thanks for getting started here, but this in complete. I've replaced it with https://github.com/GitLabPHP/Client/pull/866.

PR #844, opened 2025-09-15
Add support for fetching remote mirror(s) for projects

timdiels1:
There is no way to fetch the remote mirror(s) for a project. This PR will add it. I based myself on the protected tags, that also didn't have a test for it.

API specs: https://docs.gitlab.com/api/remote_mirrors/#list-a-projects-remote-mirrors


timdiels1:
No need to put this stale, a review needs to be done...

GrahamCampbell:
Thanks. Adjusted and merged in https://github.com/GitLabPHP/Client/pull/871.

PR #846, opened 2025-10-14
Added more options for deployments endpoint

chumanfu:
As per the docs: https://docs.gitlab.com/api/deployments/#list-project-deployments
I have added the following options for filtering:

- `updated_after`
- `updated_before`
- `finished_after`
- `finished_before`

I have also added the sort mode `finished_at` which needs to be set when filtering by `finished_after` and `finished_before`

When looking at the existing test I noticed the tests for project deployments was formatting the date incorrectly. I have updated this and updated the tests.

### DIFF
--- a/src/Api/Deployments.php
+++ b/src/Api/Deployments.php
@@ -14,12 +14,14 @@ declare(strict_types=1);
 
 namespace Gitlab\Api;
 
+use Symfony\Component\OptionsResolver\Options;
+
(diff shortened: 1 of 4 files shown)


chumanfu:
This is not stale. It just needs a maintainer to have a look.

GrahamCampbell:
This is implemented in https://github.com/GitLabPHP/Client/pull/868.

PR #848, opened 2025-11-04
fix: added missing sort parameter to allow sorting

barcigabri:


GrahamCampbell:
Thanks. I've moved this to https://github.com/GitLabPHP/Client/pull/874 and applied some corrections.

PR #851, opened 2025-12-09
Support Symfony 8

endroid:
* Options resolver

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "psr/http-client-implementation": "^1.0",
         "psr/http-factory-implementation": "^1.0",
         "psr/http-message": "^1.1 || ^2.0",
-        "symfony/options-resolver": "^5.4 || ^6.0 || ^7.0"
+        "symfony/options-resolver": "^6.4 || ^7.4 || ^8.0"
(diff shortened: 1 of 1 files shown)


endroid:
Does anyone have the time to merge this one so we can move to a stable version?

endroid:
Is anyone maintaining this repository? Otherwise it would be good to mark it as abandoned so contributors don't put unneeded effort in it.

ElGabbu:
@GrahamCampbell could we have this merged please?

composer.json
```diff
@@ -61,6 +61,12 @@
             "php-http/discovery": true
         }
     },
+    "repositories": {
```
ElGabbu, line 64: @endroid why did we need to add this repo reference?

GrahamCampbell:
Thanks, but I can't merge this in its current state due to the removal of the old version and the repositories entry.

endroid:
Thank you for updating to SF 8.0 in the meanwhile.

composer.json
endroid: Something on my side I accidentally committed :)

PR #853, opened 2026-01-14
feat(api): add job token api

Xenira:
Added endpoints as described in https://docs.gitlab.com/api/project_job_token_scopes


GrahamCampbell:
Thanks for getting started here. I've finished off this PR at https://github.com/GitLabPHP/Client/pull/877.

